PR #6, opened 2017-03-09
Spread modifier styles inside _getPopperStyle and update example

brentertz:
#### What does this PR do?

Spreads modifier styles inside _getPopperStyle and updates example.

#### Any background context you want to provide?

We ran into this need when working with a react custom Select component where we want the dropdown list width to match that of the trigger.  

The example includes a simpler scenario where we just change the background color.

#### Screenshot

screen shot 2017-03-09 at 12 32 45 pm

### DIFF
--- a/src/Popper.jsx
+++ b/src/Popper.jsx
@@ -68,7 +68,9 @@ class Popper extends Component {
     enabled:  true,
     order:    900,
     function: (data) => {
-      if ((this.state.data && !isEqual(data.offsets, this.state.data.offsets)) || !this.state.data) {
+      if (this.state.data && !isEqual(data.offsets, this.state.data.offsets) ||
+          !this.state.data
+      ) {
         this.setState({ data })
       }
     }
@@ -115,9 +117,11 @@ class Popper extends Component {
   }
 
   _getPopperStyle = () => {
+    const { data } = this.state
+
     // If Popper isn't instantiated, hide the popperElement
     // to avoid flash of unstyled content
-    if (!this._popper || !this.state.data) {
+    if (!this._popper || !data) {
       return {
         position:      'absolute',
         pointerEvents: 'none',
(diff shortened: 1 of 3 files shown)


src/Popper.jsx
```diff
@@ -139,13 +139,15 @@ class Popper extends Component {
...
     } = this.state.data.offsets.popper
+    const styles = this.state.data.styles
 
     return {
       position,
       top:        0,
       left:       0,
       transform:  `translate3d(${left}px, ${top}px, 0px)`,
       willChange: 'transform',
+      ...styles,
```
brentertz: FWIW, this is similar to how PopperJS handles things inside of its own `applyStyle` modifier.

https://github.com/FezVrasta/popper.js/blob/master/src/popper/modifiers/applyStyle.js#L53

FezVrasta: You don't need any of the above definitions if you are going to use the `styles` provided by Popper.js, the code here can be simplified.

souporserious:
I'd like to keep things natural to React where possible. Your use case sounds like it should be solved by just passing a `style` attribute to the `Popper` component. So something like `<Popper style={{ width: variableWidth, backgroundColor: 'red' }}/>`. The style prop gets merged in with the PopperJS styles [here](https://github.com/souporserious/react-popper/blob/master/src/Popper.jsx#L186-L189). Please let me know if that works or not.

FezVrasta:
The advantage of this change is that someone can use custom modifiers of Popper.js with this component. Otherwise one can only add custom modifiers that leverage on `top` and `left` properties.

souporserious:
Sorry, I did some updates. Can you rebase master please?

src/Popper.jsx
souporserious: @FezVrasta so are you saying the transform values and such do not need to be added since `styles` is used now?

PR #21, opened 2017-05-04
feat: Make Popper instance available to child context

FezVrasta:
This allows this kind of usage:

```
const Demo = ({ close, children }) =>
  <span>
    <Target>👇</Target>
    <Portal>
      <ClickOutside onClickOutside={close}>
        <Popper>
          {({ popper }) =>
            <ResizeAware onResize={popper.scheduleUpdate}>
              {children} 👋
            </ResizeAware>
          }
        </Popper>
      </ClickOutside>
    </Portal>
  </span>
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-popper",
-  "version": "0.6.0",
+  "version": "0.7.0",
   "description": "React wrapper around PopperJS.",
   "main": "lib/react-popper.js",
   "files": [
--- a/src/Popper.jsx
+++ b/src/Popper.jsx
@@ -185,7 +185,7 @@ class Popper extends Component {
         style: popperStyle,
         ['data-placement']: popperPlacement,
       }
-      return children({ popperProps, restProps })
+      return children({ popperProps, restProps, popper: this._popper })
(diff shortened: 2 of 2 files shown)


souporserious:
Thanks! Thoughts on changing it to `popperInstance` or maybe just `instance` so it's more descriptive?

FezVrasta:
Yes whatever makes sense to you it's fine.

souporserious:
Talked with my coworkers and instead of passing the instance down. I'd feel better just passing down `scheduleUpdate`. This way someone doesn't get access to methods like `destroy` and we can keep those private for React to handle.

souporserious:
I'll get an update out today with this functionality.

FezVrasta:
Okay cool, that's fine too.

lennerd:
What is the status here? I have a Popper element inside a container and I would like to update the relative position of the popper element to its target when the content of the container changes, due to a new position of the target element.

lennerd:
Sorry, the `scheduleUpdate` function is already passed to a child function. Missed that. Also had an error in my structure. Everything is working fine now.

Issue #20, opened 2017-05-03
modifier.function is deprecated, use modifier.fn


davekiss:
I kept receiving this error in my console, I think it's due to this line:

https://github.com/souporserious/react-popper/blob/2109ac0aa4210eafc668b3abe8d6e7a9ac2b2ada/src/Popper.jsx#L76

Simply changed that to fn and error went away. Can you reproduce?

holman:
Yeah, I'm getting it too; it's because of FezVrasta/popper.js#225.

FezVrasta:
Actually it's a warning, not an error 🤓

By the way, yes, that's it. One of you could send a PR to update it.

davekiss:
Sassy. I like it. PR submitted

FezVrasta:
This should be closed since #22 has been merged

souporserious:
Was going to wait until I released it on NPM :)

nicmyers:
@souporserious Was there a plan to push out a new release to NPM?

souporserious:
Sorry, pushed https://github.com/souporserious/react-popper/commit/0115651a04b27dcb80a7398f4e75f13f5a2eaf59

PR #22, opened 2017-05-04
fix: modifier.function is deprecated, use modifier.fn

davekiss:
ensures compatibility with https://github.com/FezVrasta/popper.js/pull/225

related: #20


souporserious:
Thanks! :)

PR #23, opened 2017-05-23
Pass Popper placement to Arrow

TimonVS:
This PR adds the ability to retrieve the Popper placement from within the `Arrow` component. This makes it easier to style the arrow using Styled Components.

Example:

```js
const StyledArrow = styled(Arrow)`
  ${props => props['data-popper-placement'] === 'right-start' && 'background-color: red;'}
`

const MyPopper = () => (
  <Manager>
    <Popper>
      <StyledArrow />
    </Popper>
  </Manager>
)
```

vs. current situation:

```js
const StyledPopper = styled(Popper)`
  ${props => props['data-placement'] === 'right-start' && '.popper-arrow { background-color: red; }'}
`

const MyPopper = () => (
  <Manager>
    <StyledPopper>
      <Arrow className='popper-arrow' />
    </StyledPopper >
  </Manager>
)
```

I called the prop `data-popper-placement` instead of `data-placement` because it doesn't refer to the placement of the arrow itself.

### DIFF
--- a/src/Arrow.jsx
+++ b/src/Arrow.jsx
@@ -11,11 +11,13 @@ const Arrow = (props, context) => {
     }
   }
   const arrowStyle = popper.getArrowStyle()
+  const popperPlacement = popper.getPopperPlacement()
 
   if (typeof children === 'function') {
     const arrowProps = {
       ref: arrowRef,
       style: arrowStyle,
+      ['data-popper-placement']: popperPlacement,
     }
     return children({ arrowProps, restProps })
   }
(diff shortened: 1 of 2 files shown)


souporserious:
Sorry, I just got a chance to look at this. You can do this with just CSS :) as long as you can use the `&` selector you should be able to write this:

```
const StyledArrow = styled(Arrow)`
  [data-placement="right-start"] & {
    background-color: red;
  } 
`
```

TimonVS:
That works perfectly! Thanks a lot @souporserious :)

PR #26, opened 2017-06-14
Fix needed to properly work with Popper.js 1.10.1

FezVrasta:
Popper.js 1.10.1 added two new options to the new `computeStyle` modifier, now the poppers can be positioned using a different offset origin (`bottom: 0` or `left: 0` instead of the usual `top` and `left`).

To make the new options work properly you have to get rid of all the manual styles computation and simply take what Popper.js provides.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-popper",
-  "version": "0.6.6",
+  "version": "0.6.7",
   "description": "React wrapper around PopperJS.",
   "main": "lib/react-popper.js",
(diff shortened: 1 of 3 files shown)


FezVrasta:
@souporserious sorry for pinging you again after just few hours, but this is blocking for some stuff we are working on and we need this as soon as possible please 🙏

souporserious:
@FezVrasta done! Sorry I missed that, I've been swamped these last few days 😓

Issue #32, opened 2017-07-06
Uncaught at dispatch TypeError: Cannot read property 'scheduleUpdate' of undefined


guillaumepotier:
Hi,

I'm facing a console bug with Popper [at that line](https://github.com/souporserious/react-popper/blob/b3e0f32813dac0779f06d569ef456af29551add3/src/Popper.jsx#L57)

Dunno why, and things seems to work smoothly despite that error.

Should not we test `this._popper` before, like done above in the file?

souporserious:
Should be fixed with https://github.com/souporserious/react-popper/pull/38 thank you 🙏

PR #38, opened 2017-07-12
Fix scheduleUpdate call if this._popper does not exist

guillaumepotier:
Refs #32

Defensive PR to avoid error if `this._popper` not defined. Should inspect to understand why not defined in the first place, but really needed to fix it for my project.

Merge if it helps =)

Best


souporserious:
Thank you! 🙏

PR #45, opened 2017-08-01
Fix missing data-x-out-of-boundaries attribute

kylegach:
### Current behavior
No support for https://popper.js.org/popper-documentation.html#modifiers..hide.

### Expected behavior
Full support for `modifiers.hide`, specifically that the `data-x-out-of-boundaries` attribute is passed along.

### Changes
Passes along the `data-x-out-of-boundaries` attribute in a similar manner to `data-placement`.

### Why
We needed this in our app. More importantly, we saw no reason a wrapper around Popper shouldn’t support this.

### How this has been tested
Confirmed `build` still works. Confirmed expected example behavior when running `dev` after modifying the Popper config like so:

```js
const modifiers = {
  customStyle: {
    // ...unchanged
  },
  preventOverflow: {
    boundariesElement: 'viewport',
  },
  hide: {
    enabled: true,
  },
}
```

### DIFF
--- a/src/Popper.jsx
+++ b/src/Popper.jsx
@@ -147,6 +147,10 @@ class Popper extends Component {
     return !!this.state.data ? this.state.data.placement : undefined
   }
 
+  _getPopperHide = () => {
+    return !!this.state.data && this.state.data.hide ? '' : undefined
+  }
+
   _getArrowStyle = () => {
     if (!this.state.data || !this.state.data.offsets.arrow) {
       return {}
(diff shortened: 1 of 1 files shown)


kylegach:
Hi, @souporserious. Have you had a chance to consider this? We still need it within our app, and would love to not have to rely on my fork.

FezVrasta:
You may create a custom modifier to add this missing piece of logic. It's fairly easy

FezVrasta:
Or directly use the `data` object to obtain this info

souporserious:
Sorry, I've been super busy lately and just getting around to this. Looks good! Thanks!

kylegach:
Thanks, Travis!

PR #46, opened 2017-08-02
Update to use latest typings from popper.js

rajington:
The typings in popper.js were [updated recently](https://github.com/FezVrasta/popper.js/pull/367#issuecomment-319449475), causing issues. I believe this fix is sufficient and so does the [update's author](https://github.com/FezVrasta/popper.js/pull/367#issuecomment-319674674) but I would like to verify with someone if this is the right step.

### DIFF
--- a/react-popper.d.ts
+++ b/react-popper.d.ts
@@ -1,6 +1,6 @@
 declare module "react-popper" {
   import * as React from "react";
-  import * as PopperJS from "popper.js";
+  import PopperJS from "popper.js";
 
   interface IRestProps {
     restProps: {


madebyherzblut:
I do have the same problem and your fix works. Would love to see this merged :)

adidahiya:
Please hold off on this PR; see my comment here: https://github.com/FezVrasta/popper.js/pull/367#issuecomment-322680526

@madebyherzblut In the meantime, you should enable the `--allowSyntheticDefaultImports` compiler flag.

rajington:
I'll close this one but a link to a PR with a reverse commit in popper.js would be good

explicitcall:
@adidahiya `--allowSyntheticDefaultImports` doesn't fix this for me, see https://github.com/FezVrasta/popper.js/pull/367#issuecomment-324120105. Basically latest version of react-popper doesn't pass typechecks with TypeScript.

PR #51, opened 2017-08-17
Update how react-popper.d.ts imports PopperJS

oeb25:
`popper.js`'s TypeScript definition now exports default, so it's now throwing an error when importing with `* as PopperJS from "popper.js"`. This is a small bug, but it prevents anyone using TypeScript from building their application.

### DIFF
--- a/react-popper.d.ts
+++ b/react-popper.d.ts
@@ -1,6 +1,6 @@
 declare module "react-popper" {
   import * as React from "react";
-  import * as PopperJS from "popper.js";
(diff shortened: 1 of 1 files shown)


pmoleri:
Hi @souporserious, merging this PR is important to anyone using typescript.
It would also make sense to upgrade dependency `popper.js` to `^1.12.1` given that the new typings are in versions `>= 1.11.1`.

souporserious:
Sorry, I've been super busy lately. I'll get this out today, thanks!

PR #55, opened 2017-08-29
Subscribe to Popper updates in the Arrow component

### LINKED ISSUE #54: arrowStyle props not being passed to Arrow component
I'm trying to have an arrow next to a popover that moves with the popover. This seems to be supported in vanilla Popper.js, and the relevant code [seems to be present](https://github.com/souporserious/react-popper/blob/b3e0f32813dac0779f06d569ef456af29551add3/src/Arrow.jsx#L13) in `react-popper` to pass the style computed from Popper down into the `Arrow` component, however in my app, I see `style={}` on that component.

I used some logs and the debugger to try to narrow down the issue: I think Popper is mounting and implicitly `Arrow` needs to re-render to consider the new props passed in via style. However, it doesn't rerender:

```
15:57:51.482 Popper.js:112 getChildContext Popper {props: {…}, context: {…}, refs: {…}, updater: {…}, state: {…}, …}
15:57:51.498 Popper.js:96 _getArrowStyle Popper {props: {…}, context: {…}, refs: {…}, updater: {…}, state: {…}, …}
15:57:51.522 Popper.js:144 _updatePopper Popper {props: {…}, context: {…}, refs: {…}, updater: {…}, state: {…}, …}context: {popperManager: {…}}props: {placement: "right", children: {…}, component: "div", eventsEnabled: true, modifiers: {…}}refs: {}state: {data: {…}}updater: {isMounted: ƒ, enqueueCallback: ƒ, enqueueCallbackInternal: ƒ, enqueueForceUpdate: ƒ, enqueueReplaceState: ƒ, …}_arrowNode: div.settings_editor_arrow_getArrowStyle: ƒ ()_getPopperPlacement: ƒ ()_getPopperStyle: ƒ ()_getTargetNode: ƒ ()_node: div_popper: Popper {options: {…}, state: {…}, reference: div, scheduleUpdate: ƒ, update: ƒ, …}_reactInternalInstance: ReactCompositeComponentWrapper {_currentElement: {…}, _rootNodeID: 0, _compositeType: 0, _instance: Popper, _hostParent: ReactDOMComponent, …}_setArrowNode: ƒ (node)_updateStateModifier: {enabled: true, order: 900, fn: ƒ}isMounted: (...)replaceState: (...)__proto__: ReactComponent
15:57:51.524 Popper.js:153 creating popper
15:57:51.535 Popper.js:112 getChildContext Popper {props: {…}, context: {…}, refs: {…}, updater: {…}, state: {…}, …}
15:57:51.541 Popper.js:112 getChildContext Popper {props: {…}, context: {…}, refs: {…}, updater: {…}, state: {…}, …}
```

My best understanding of the issue is that React doesn't know that `componentDidMount()` implicitly means that `Arrow` must rerender, thus Arrow is stuck behind.

If I manually click on `MyArrowComponent` instance in the React devtools and type  `$r.forceUpdate()` I can get it to have the correct behavior.

darknoon:
I'm not an expert in how to do subscriptions in React, but this was my best guess.

Happy to look at alternates if there is a better way to fix this bug! 🔎🐛

Fixes #54

### DIFF
--- a/src/Arrow.jsx
+++ b/src/Arrow.jsx
@@ -1,40 +1,66 @@
 import React, { Component, createElement } from 'react'
 import PropTypes from 'prop-types'
 
-const Arrow = (props, context) => {
-  const { component = 'span', innerRef, children, ...restProps } = props
-  const { popper } = context
-  const arrowRef = node => {
-    popper.setArrowNode(node)
-    if (typeof innerRef === 'function') {
-      innerRef(node)
+class Arrow extends Component {
+  // State is updated by the containing Popper component when popper.js updates
+  state = { arrowStyle: {} }
+
+  // Subscribe on mount
+  componentDidMount() {
+    const { props, context } = this
+    const { popper: { subscribe } } = context
+    subscribe(this._update)
+  }
+
+  // Unsubscribe on unmount
+  componentWillUnmount() {
+    const { props, context } = this
+    const { popper: { unsubscribe } } = context
+    unsubscribe(this._update)
+  }
+
+  _update = data => {
+    const { top, left } = data.offsets.arrow
+    this.setState({ arrowStyle: { top, left } })
+  }
+
+  render() {
+    const { props, context, state } = this
+    const { component = 'span', innerRef, children, ...restProps } = props
+    const { popper } = context
+    const { arrowStyle } = state
+
+    const arrowRef = node => {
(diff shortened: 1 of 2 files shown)


souporserious:
This _should_ work since Arrow has to be a child of Popper, and it sends the data down. Please feel free to open this back up if you find another bug.

PR #61, opened 2017-10-04
Updates TypeScript definition for IPopperChildrenProps

jamiewinder:
Both `restProps` and `scheduleUpdate` properties were missing from the render function parameter type.


souporserious:
Thank you 🙏

PR #62, opened 2017-10-06
make Popper child match parent (Target) width

moonith:
When used with some portal library (mounting element to separate body tree) the width of popper does not match target. This pr solves it. You can still manually set width of Popper child if needed.

### DIFF
--- a/src/Popper.jsx
+++ b/src/Popper.jsx
@@ -147,10 +147,10 @@ class Popper extends Component {
     }
 
     const { top, left, position } = data.offsets.popper
-
     return {
       position,
       ...data.styles,
+      ...{width: data.offsets.reference.width}
     }
   }
 


FezVrasta:
I think you should apply this logic in your own code, this isn't the desired behavior most of the times.

moonith:
Yes, they way i did it is a little naive, but I think there should be a simple way to retrive target dimensions in Popper component. Maybe in popperProps?

FezVrasta:
It's not that easy to get correct size because it could change.

Your best bet is to use react-resize-aware and wire everything together

souporserious:
Sorry, I'd rather not introduce this logic here. You can use react-resize-aware as suggested above or even something like a [ResizeObserver](https://github.com/que-etc/resize-observer-polyfill) to maintain proper widths.

PR #67, opened 2017-10-25
Fix URL to Popper modifiers arrow documentation

richardvenneman:
Thanks for this lib!

Coming message should be pretty self-explanatory :)


souporserious:
Thank you 🙏

PR #70, opened 2017-11-12
feat: Make Manager return a fragment by default

### LINKED ISSUE #69: Major version with React 16 as primary React version?
I was thinking that we may ship a breaking change as major version that switches the value of the `tag` of `Manager` to `false` by default.

This will make the library work with the new wrapless-components of React 16 but will still allow to fallback to the old behavior of a wrapping element if user is using an older version.

What's your take on this?

FezVrasta:
BREAKING CHANGE: The default value of the `tag` prop of`Manager` has changed from `"div"` to `false`.

This aims to close #69


src/Manager.jsx
```diff
@@ -11,7 +11,9 @@ class Manager extends Component {
   }
 
   static defaultProps = {
-    tag: 'div',
+    // Using React 15 or lower? Remember to set this to `div` or
+    // another valid HTML tag!
+    tag: false,
```
FezVrasta: I was also thinking to automatically detect the React version and set the default value accordingly, but I'm not sure it's a good idea.

piecyk: @FezVrasta IMHO this is correct, just add an note to change-log/release about this and should be good... Automatically detect the React version just to set an boolean is always an bad idea! 

Overall awesome work with this 👏

jaxyeh:
@FezVrasta maybe you guys want to consider using [react-aux](https://github.com/gajus/react-aux) library for those legacy React version so that this component API behaves the same regardless of which React version. I know it's breaking changes for those who likes to keep `tag` props, but I figured it might be good idea to go forward with "same" API design. You may fork the 0.8.x ("legacy") branch elsewhere. Just sharing thoughts...

FezVrasta:
It doesn't make much sense to have a fallback from fragment to HTML tag, they have two different behaviors that could confuse the user IMHO

FezVrasta:
Closing in favor of https://github.com/souporserious/react-popper/pull/117

PR #71, opened 2017-11-13
Pass the placement directly to the Arrow

sgenoud:
This is useful for `Arrow` implemented with `styled-components` for instance (which can style the arrow based on the attribute instead of having to set some classes.

### DIFF
--- a/src/Arrow.jsx
+++ b/src/Arrow.jsx
@@ -11,16 +11,19 @@ const Arrow = (props, context) => {
     }
   }
   const arrowStyle = popper.getArrowStyle()
+  const popperPlacement = popper.getPlacement()
 
   if (typeof children === 'function') {
     const arrowProps = {
       ref: arrowRef,
       style: arrowStyle,
+      ['data-placement']: popperPlacement,
     }
     return children({ arrowProps, restProps })
   }
 
   const componentProps = {
+    ['data-placement']: popperPlacement,
     ...restProps,
     style: {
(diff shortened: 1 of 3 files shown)


FezVrasta:
I'm quite sure styled-components allow something like `[data-placement="bottom"] &`, but I see your point here.

sgenoud:
I must admit that I did not think of building it with `[data-placement="bottom"] &`. 

Generally it might be helpful to expose the placement directly to the arrow as you might want to do trickier logic within your arrow.

Would you like that I add some documentation for this feature?

FezVrasta:
It's up to @souporserious 😊

src/Arrow.jsx
```diff
@@ -11,16 +11,19 @@ const Arrow = (props, context) => {
     }
   }
   const arrowStyle = popper.getArrowStyle()
+  const popperPlacement = popper.getPlacement()
 
   if (typeof children === 'function') {
     const arrowProps = {
       ref: arrowRef,
       style: arrowStyle,
+      ['data-placement']: popperPlacement,
```
kaifaust: How about naming the prop `placement`?

souporserious:
Sorry for the late reply, this has been discussed [here](https://github.com/souporserious/react-popper/pull/23#issuecomment-304763886). I'd rather not add more to the library if we don't need to. Thank you though, I appreciate it.

PR #83, opened 2018-01-23
fix: Updated examples

greg-a-smith:
Refactored examples into separate components for clarity. Added an example passing custom modifiers to the Popper component and to demonstrate the out-of-boundaries hiding.


souporserious:
Thank you! 🙏

PR #89, opened 2018-02-06
chore: update & clean dependencies

Justkant:
Hi, I wanted to only move react-transition-group to devDependencies but in the end I updated everything I could think of. Tell me if you need some things to remain the same (e.g., sass, react-travel) :v:

chore:
- add `eslint` with `standard` & `prettier`
- add `husky` & `lint-staged` to run linters on git staged files
- exclude `node_modules` with the babel-loader
- use `prepare` instead of `prepublish`
- clean unused dependencies
- example: remove `sass` usage (not really used)
- example: replace `react-travel` with `react-portal` (react 16 support)
- example: move `react-transition-group` to devDependencies

refactor:
- using `prettier`


Justkant:
There is maybe some problem with the way babel compile using babel-preset-env, i'm currently looking for the best way to have a good library build system & keep the same compatibility as before

souporserious:
Dangit 🤦‍♂️ I didn't realize `react-transition-group` was added to deps. That's what I get for working late 😅. Thanks for the changes! I'll try and look through them today and get it merged in.

souporserious:
This looks great! Thank you so much for the contribution 🙏

Justkant:
Thanks :v:
I'm still gonna look because I think the build is not right for all browsers environments ([example](https://stackblitz.com/edit/react-nxup9u))
[this PR Manager.js](https://unpkg.com/@kant/react-popper@0.8.0/lib/Manager.js)
[current Manager.js](https://unpkg.com/react-popper@0.8.0/lib/Manager.js)

jaxyeh:
Giving you guys heads up that this 0.8.1 release broke our environment because it's using ES6 import/export and we had to revert back to 0.8.0. See [before](https://unpkg.com/react-popper@0.8.0/lib/react-popper.js) and [current](https://unpkg.com/react-popper@0.8.1/lib/react-popper.js).

Justkant:
Yep actually looking into it

Justkant:
@jaxyeh It should be ok with #90

souporserious:
Thanks @Justkant! Published under `0.8.2` 🎉

PR #91, opened 2018-02-07
Move `prop-types` as peerDependency

pascalduez:
Hi,

- Moved `prop-types` as a `peerDependency`, similar to how we rely on `react`, `react-dom`, to prevent possible duplicate versions in app bundles.
- Simplified a bit the `react`, `react-dom` `peerDependencies` declarations by using ranges.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,12 +42,12 @@
     "prepare": "npm run build"
   },
   "peerDependencies": {
+    "prop-types": "15.x",
     "react": "0.14.x || ^15.0.0 || ^16.0.0",
     "react-dom": "0.14.x || ^15.0.0 || ^16.0.0"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


TheSharpieOne:
https://github.com/facebook/prop-types#how-to-depend-on-this-package

> For libraries, we also recommend leaving it in dependencies:
> ```
>   "dependencies": {
>     "prop-types": "^15.5.7" 
>   },
>   "peerDependencies": {
>     "react": "^15.5.0" 
>   }
> ```
> Note: there are known issues in versions before 15.5.7 so we recommend using it as the minimal version.

Justkant:
@FezVrasta do you think we can close this PR now ?

But I like the simplified declarations using ranges for peerDependencies, I'll do a PR later for that :v:

PR #93, opened 2018-02-22
fix: build exports

Justkant:
bump dev dependencies

We are back to v0.8.0 for `react-popper.js`
[react-popper.js@v0.8.0](https://unpkg.com/react-popper@0.8.0/lib/react-popper.js)
[react-popper.js@v0.8.1](https://unpkg.com/react-popper@0.8.1/lib/react-popper.js)
[react-popper.js@v0.8.2](https://unpkg.com/react-popper@0.8.2/lib/react-popper.js)

The problem was that `export Manager from './Manger'` is transformed differently than `export { Default as Manager } from './Manager'` by babel

### DIFF
--- a/src/react-popper.js
+++ b/src/react-popper.js
@@ -1,4 +1,4 @@
-export { default as Manager } from './Manager'
-export { default as Target } from './Target'
-export { default as Popper } from './Popper'
-export { default as Arrow } from './Arrow'
+export Manager from './Manager'
+export Target from './Target'
+export Popper from './Popper'
+export Arrow from './Arrow'
--- a/package.json
+++ b/package.json
@@ -4,10 +4,13 @@
   "description": "React wrapper around PopperJS.",
   "main": "lib/react-popper.js",
   "types": "react-popper.d.ts",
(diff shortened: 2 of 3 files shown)


supergibbs:
Awesome thanks!

supergibbs:
Any update on merging this and cutting a release?

src/react-popper.js
```diff
@@ -1,4 +1,4 @@
-export { default as Manager } from './Manager'
-export { default as Target } from './Target'
-export { default as Popper } from './Popper'
-export { default as Arrow } from './Arrow'
+export Manager from './Manager'
```
TheSharpieOne: Is this valid syntax?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export
https://stackoverflow.com/q/41139740/1873485

Justkant: Only with babel for the moment: https://babeljs.io/docs/plugins/transform-export-extensions/

lednhatkhanh:
When is the fix going to be released?

FelixHenninger:
Thank you! 💛

PR #94, opened 2018-02-23
fix: minified umd build without prop-types

### LINKED ISSUE #80: Remove Prop-Types from Production Build
It would be nice if PropTypes wasn't in the production build in any fashion. I believe this could be achieved by utilizing the following plugin:

https://www.npmjs.com/package/babel-plugin-transform-react-remove-prop-types

This would allow using the library without having to include the prop-types library as a pre-requisite in production.

Justkant:
Fix #80 

I had a hard time finding that the `cross-env` usage for the minified version was one of the problems.

`cross-env NODE_ENV=production && cross-env TARGET=minify webpack` is not setting `NODE_ENV` for webpack

`cross-env NODE_ENV=production TARGET=minify webpack` is the proper way to use `cross-env`

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,8 +1,5 @@
 {
-  "presets": [
-    ["env"],
-    "react"
-  ],
+  "presets": ["env", "react"],
   "plugins": [
     "transform-object-rest-spread",
     "transform-class-properties",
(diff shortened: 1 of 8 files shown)


Justkant:
Just rebased on master to remove all conflicts

FezVrasta:
Sorry I'm closing this in favor of #109

Justkant:
No problem, nice to see rollup in use here :v:

PR #97, opened 2018-03-02
Moved import statements out of module declaration

daxtero:
In my opinion the imports of react and popper.js have to be outside of "react-popper" module declaration.

This will result in problems for example with my fuse box dev server.

### DIFF
--- a/react-popper.d.ts
+++ b/react-popper.d.ts
@@ -1,6 +1,7 @@
+import * as React from "react";
+import PopperJS from "popper.js";
+
 declare module "react-popper" {
-  import * as React from "react";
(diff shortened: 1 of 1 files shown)


react-popper.d.ts
```diff
@@ -1,6 +1,7 @@
+import * as React from "react";
+import PopperJS from "popper.js";
```
rolandjitsu: I would actually use `import * as PopperJs from 'popper.js'` as with the new version of Popper the TS compiler seems to complain, see https://github.com/souporserious/react-popper/issues/101.

PR #100, opened 2018-03-07
feat: Add onFlip handler

buberdds:
Maybe I didn't catch everything, but when there's not enough space to render popper at bottom, it will be placed at top for instance, and this is great and desired. Does `Target` is aware of `Popper` placement set by popperjs when flip occurs?

PR adds a callback to get a flipped and a placement values in a parent component. Using an object returned via a callback it is possible to adjust classes or styles. Depends on placement `Target` and `Popper` may look differently. I think `onFlip` is a small replacement for onUpdate.

If this is some kind of edge case or it's totally not needed here let me know!

have a good one!


FezVrasta:
This would get called on each update, not only when the placement flipped, right?

buberdds:
Thanks for a feedback. Now it's called only when flipped state changed.

src/Popper.jsx
```diff
@@ -38,15 +39,22 @@ class Popper extends Component {
     }
   }
 
-  componentDidUpdate(lastProps) {
+  componentDidUpdate(lastProps, lastState) {
+    const { children, eventsEnabled, onFlip, placement } = this.props
+    const { data: { placement: currentPlacement, flipped } } = this.state
+
+    if (lastState.data && lastState.data.flipped !== flipped) {
+      onFlip && onFlip({ placement: currentPlacement, flipped })
```
TheSharpieOne: should move the `onFlip &&` check to the if condition: `onFlip && lastState.data && lastState.data.flipped !== flipped`

buberdds:
thx @TheSharpieOne. 
I found one additional case, when popper is not visible on init and later when an open state change it is mounted with flipped state set to true.

src/Popper.jsx
```diff
@@ -38,15 +39,26 @@ class Popper extends Component {
     }
   }
 
-  componentDidUpdate(lastProps) {
+  componentDidUpdate(lastProps, lastState) {
+    const { children, eventsEnabled, onFlip, placement } = this.props
+    const { data: { placement: currentPlacement, flipped } } = this.state
```
Justkant: If `data` is `undefined` or `null` we could have a `TypeError`
You can use `const { data: { placement: currentPlacement, flipped } = {} } = this.state`
or init `data` in the state object.

src/Popper.jsx
```diff
@@ -38,15 +39,26 @@ class Popper extends Component {
...
-      lastProps.placement !== this.props.placement ||
-      lastProps.eventsEnabled !== this.props.eventsEnabled
+      lastProps.placement !== placement ||
+      lastProps.eventsEnabled !== eventsEnabled
     ) {
       this._destroyPopper()
       this._createPopper()
     }
-    if (lastProps.children !== this.props.children) {
+    if (lastProps.children !== children) {
```
Justkant: You could destructure `lastProps` & `lastState` like `this.props` & `this.state` in the function params

FezVrasta:
I wonder if we still need this now that we have the new API?

I think we can do something like this instead?

```
<Popper placement={initialPlacement}>
  {({ placement, ...props }) =>
     <div {...props}>
	   {initialPlacement !== placement ? 'flipped' : 'not flipped'}
	  </div>
  }
</Popper>
```

Justkant:
Yeah the onFlip handler seems easier to understand and use, but there's no need to have it with the new API.
Some examples of this with the new API in the README would help newcomers.

FezVrasta:
Please feel free to either update this PR to use he new code-base or add an example to the readme.

buberdds:
@FezVrasta, I just briefly tested the new API and I think it resolve my case perfectly. In this case I don't see a value in this PR. Thank you for your great work, kudos!

PR #103, opened 2018-03-12
fix for compilation error 'Cannot find namespace PopperJS'

xaviergonz:
When you compile the current code with TS 2.6.2 (didn't try others, and at least with allowSyntheticDefaultImports set to false) it gives the following errors:
```
build/0/compileTypescript/0  node_modules/react-popper/react-popper.d.ts(30,27): error TS2503: Cannot find namespace 'PopperJS'.
build/0/compileTypescript/0
build/0/compileTypescript/0  30       ["data-placement"]: PopperJS.Placement;
build/0/compileTypescript/0                               ~~~~~~~~
build/0/compileTypescript/0
build/0/compileTypescript/0
build/0/compileTypescript/0  node_modules/react-popper/react-popper.d.ts(38,17): error TS2503: Cannot find namespace 'PopperJS'.
build/0/compileTypescript/0
build/0/compileTypescript/0  38     modifiers?: PopperJS.Modifiers;
build/0/compileTypescript/0                     ~~~~~~~~
build/0/compileTypescript/0
build/0/compileTypescript/0
build/0/compileTypescript/0  node_modules/react-popper/react-popper.d.ts(39,17): error TS2503: Cannot find namespace 'PopperJS'.
build/0/compileTypescript/0
build/0/compileTypescript/0  39     placement?: PopperJS.Placement;
```
This little fix fixed that.

PR #104, opened 2018-03-13
Update Popper Typings

Ritorna:
As in version 1.14.0 of popper.js the Export typings had been changed, which is now using Default Exports. 
See the changes as in the Commit of 26th February: 
https://github.com/FezVrasta/popper.js/commit/3315a6823f1685684fccf45f3fde407f1ca3716c#diff-01613ef3d5a6b03e140f807e6a8379bb


Ritorna:
Closed as already mentioned in #103

PR #105, opened 2018-03-13
fix es6 imports

pravdomil:
in enviroments that do not support es6 imports, react-poper fails to run, this PR defines es6 modules under "module" property, read https://github.com/rollup/rollup/wiki/pkg.module


pravdomil:
can you release new version so that it's fixed on npm?

PR #108, opened 2018-03-19
chore: update dependencies

Justkant:
- also cleaned the lockfile (`yarn check` was returning a lot of warnings)
- and added `yarn-error.log` to the `.gitignore` file


FezVrasta:
Thanks!

PR #116, opened 2018-03-30
Fix pass of custom options for arrow modifier

langovoi:


langovoi:
Thanks @FezVrasta!

When this will be published to npm?

FezVrasta:
0.9.1

PR #117, opened 2018-03-30
feat: New react-popper API

FezVrasta:
New API is render props based and it looks like this.
I also replaced PropTypes with Flow types.

You can play with it running `yarn demo`, or use it in your project installing `react-popper@1.0.0-alpha.1`.

Feedbacks welcome!

```jsx
<Manager>
  <Target>
    {({ getTargetRef }) => (
      <div
        ref={getTargetRef}
        className="target"
      >
        Target
      </div>
    )}
  </Target>
  <Popper placement={this.state.placement}>
    {({ getPopperRef, style, placement }) => (
      <div
        data-placement={placement}
        ref={getPopperRef}
        className="popper"
        style={style}
      >
        Popper
        <Arrow>
          {({ getArrowRef, style }) => (
            <div
              ref={getArrowRef}
              className="popper__arrow"
              style={style}
            />
          )}
        </Arrow>
      </div>
    )}
  </Popper>
</Manager>
```

**edit**:
cool, I just tried this new code with React 15, it actually works. You just have to wrap `Target` and `Popper` in a `div`:

```
<Manager>
  <div>
    <Target />
    <Popper />
  </div>
</Manger>
```

# To do:

- [x] Update TypeScript definitions (need help!)
- [x] Ship Flow types definitions
- [x] Add tests


FezVrasta:
I updated the TypeScript d.ts file but it's failing because `HTMLElement can't be assigned to HTMLButtonElement`... Any clue on how to tell TS to accept ANY HTML element?

souporserious:
This is looking great! I've been meaning to look at a newer API similar to what you have to solve https://github.com/souporserious/react-popper/pull/70. How do you feel about moving all of the render props up into one component? So something like:

```
<Popper>
  { ({ getTargetProps, getPopperProps }) =>
    <Fragment>
      <div {...getTargetProps()}>Target</div>
      <div {...getPopperProps()}>Popper</div>
    </Fragment>
  }
</Popper>
```

I was thinking this could be a little simpler and get rid of all of the context stuff, but maybe the API is fine the way it is 😅.

FezVrasta:
Thanks for the feedback Travis.

What would it be the advantage of having functions that return props instead of objects?

About the single component API, I think it's a cool idea to reduce boilerplate, I hope it's not too different from what people are used to, what do you think?

FezVrasta:
@souporserious may you enable Travis CI tests?

souporserious:
I guess I was thinking `getPopperProps()` would generate a unique id or something so you could have multiple poppers with one target. Might be better to just keep them in components since it's already setup for that.

Pertaining to the Travis CI tests, do I just need to signup for an account and add it in the settings section of this repo? 😇 sorry, I've never setup CI testing 💩

FezVrasta:
Yes you should find the instructions in the settings of the repo.

Good point about multiple Poppers, at the moment we lost this ability with the new-new API 🤔

FezVrasta:
I guess we may have something like this then, so that we reduce the boilerplate but we keep the ability to have multiple poppers into a single Manager?

```
<Manager>
  <Target>
    {({ getTargetRef }) => (
      <div
        ref={getTargetRef}
        className="target"
      >
        Target
      </div>
    )}
  </Target>
  <Popper>
    {({ popperProps, arrowProps }) => (
      <div
        data-placement={popperProps.placement}
        ref={popperProps.getRef}
        style={popperProps.style}
      >
        Popper
        <div
          ref={arrowProps.getRef}
          style={arrowProps.style}
        />
      </div>
    )}
  </Popper>
</Manager>
```

kentcdodds:
Here's an explanation for why prop getters are better: “How to give rendering control to users with prop getters” https://blog.kentcdodds.com/how-to-give-rendering-control-to-users-with-prop-getters-549eaef76acf

FezVrasta:
Thanks for the reply @kentcdodds! The challenge here is to connect the "popper" and "arrow" elements together.

I can have a `getPopperProps()` prop getter that allows me to create multiple poppers into a single "context", but if I want to associate an arrow to a specific popper things get tricky.

```
<Popper>
  ({ getReferenceProps, getPopperProps }) => 
	<Fragment>
		<div {...getReferenceProps()} />
		<div {...getPopperProps()}>
			<div>arrow?</div>
		</div>
		<div {...getPopperProps()}>
			<div>arrow?</div>
		</div>
	</Fragments>
</Popper>
```

souporserious:
Yeah I think [this](https://github.com/souporserious/react-popper/pull/117#issuecomment-377796224) is perfect. If you want to go the other route, you could look at [how it's done in react-measure now](https://github.com/souporserious/react-measure/blob/rewrite/src/Measure.js#L33). Basically we keep a list of each call to `bind` and reuse the stored functions, but in react popper this would require naming each popper call which could be more tedious over just using a component that can take care of it. So I guess what I'm saying is I think what you've originally proposed is great 😇.

package.json
```diff
@@ -59,34 +71,58 @@
...
     "babel-plugin-external-helpers": "^6.22.0",
+    "babel-plugin-transform-class-properties": "^6.24.1",
     "babel-plugin-transform-react-remove-prop-types": "^0.4.13",
     "babel-preset-env": "^1.6.1",
     "babel-preset-react": "^6.24.1",
     "babel-preset-stage-2": "^6.24.1",
     "chokidar": "^1.6.1",
     "create-styled-element": "^0.4.0",
+    "enzyme": "^3.3.0",
+    "enzyme-adapter-react-16": "^1.1.1",
```
langovoi: Look at this: https://github.com/i18next/react-i18next/pull/414
Maybe use multiple versions of React for tests?

FezVrasta: Cool thanks, I wasn't aware of this helper.

FezVrasta: Going to wait for Enzyme to release a new version of the react-16 adapter to support the new Context API then I'll implement this.

souporserious:
Great work! 🎉 I think this API is wonderful. Thank you for handling everything 🙏.

FezVrasta:
Thanks! I'm going to keep this beta on the `@next` tag for a bit to see if people report any problem, if everything is fine I'll move it to stable.

Justkant:
I'll give it a try in the next days. Nice work 👏

src/Manager.js
```diff
@@ -1,45 +1,35 @@
...
   render() {
-    const { tag, children, ...restProps } = this.props
-    if (tag !== false) {
-      return createElement(tag, restProps, children)
-    } else {
-      return children
-    }
+    return (
+      <ManagerContext.Provider
+        value={{
```
tyscorp: Doing this will generate a new object every render. Is this intended?

souporserious: Oops! I made this mistake on react-measure. More info [here](https://twitter.com/dan_abramov/status/980950981981609984).

FezVrasta: How am I supposed to put the `getReferenceRef` function in the state? 😔

PR #129, opened 2018-04-05
perf: Use babel-plugin-transform-runtime to reduce size

FezVrasta:
With this change, the `cjs` and `es` builds will refer to `babel-runtime` package to import the needed polyfills instead of inlining them on each file.

This adds `babel-runtime` as additional dependency, but I don't think we should be worried because the final size is going to be smaller than when we didn't rely on it.

Example of transpiled file before:
```
var _extends = Object.assign || function (target) { for (var i = 1; i < arguments.length; i++) { var source = arguments[i]; for (var key in source) { if (Object.prototype.hasOwnProperty.call(source, key)) { target[key] = source[key]; } } } return target; };

var _createClass = function () { function defineProperties(target, props) { for (var i = 0; i < props.length; i++) { var descriptor = props[i]; descriptor.enumerable = descriptor.enumerable || false; descriptor.configurable = true; if ("value" in descriptor) descriptor.writable = true; Object.defineProperty(target, descriptor.key, descriptor); } } return function (Constructor, protoProps, staticProps) { if (protoProps) defineProperties(Constructor.prototype, protoProps); if (staticProps) defineProperties(Constructor, staticProps); return Constructor; }; }();

function _classCallCheck(instance, Constructor) { if (!(instance instanceof Constructor)) { throw new TypeError("Cannot call a class as a function"); } }

function _possibleConstructorReturn(self, call) { if (!self) { throw new ReferenceError("this hasn't been initialised - super() hasn't been called"); } return call && (typeof call === "object" || typeof call === "function") ? call : self; }

function _inherits(subClass, superClass) { if (typeof superClass !== "function" && superClass !== null) { throw new TypeError("Super expression must either be null or a function, not " + typeof superClass); } subClass.prototype = Object.create(superClass && superClass.prototype, { constructor: { value: subClass, enumerable: false, writable: true, configurable: true } }); if (superClass) Object.setPrototypeOf ? Object.setPrototypeOf(subClass, superClass) : subClass.__proto__ = superClass; }

import React, { Component } from "react";
import createContext from "create-react-context";
```

and after:

```
import _extends from "babel-runtime/helpers/extends";
import _Object$getPrototypeOf from "babel-runtime/core-js/object/get-prototype-of";
import _classCallCheck from "babel-runtime/helpers/classCallCheck";
import _createClass from "babel-runtime/helpers/createClass";
import _possibleConstructorReturn from "babel-runtime/helpers/possibleConstructorReturn";
import _inherits from "babel-runtime/helpers/inherits";
import React, { Component } from "react";
import createContext from "create-react-context";
```

For sake of simplicity, the `browser` (umd) bundle is not going to require the `babel-runtime` dependency.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,6 @@
 {
   "presets": [["env", { "modules": false }], "stage-1", "react"],
-  "plugins": ["transform-class-properties"],
+  "plugins": ["transform-class-properties", "transform-runtime"],
   "env": {
     "test": {
       "presets": ["env", "stage-1", "react"]
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "react-dom": "0.14.x || ^15.0.0 || ^16.0.0"
   },
   "dependencies": {
+    "babel-runtime": "6.x.x",
     "create-react-context": "^0.2.1",
     "popper.js": "^1.14.1",
     "warning": "^3.0.0"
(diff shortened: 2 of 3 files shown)


souporserious:
This looks great! 👍🎉This was just added to react-measure as well to reduce the bundle.

PR #132, opened 2018-04-05
Propagate Popper.js updates through context

### LINKED ISSUE #121: Feature: Propagate scheduleUpdate through nested poppers
When `scheduleUpdate` is called on a popper, it should call `scheduleUpdate` on all nested poppers. We can do this through context so it works through portals, too.

I'd be happy to take this on.

This would fix this issue: https://github.com/palantir/blueprint/issues/2031

tyscorp:
Fixes #121

There are still a few flow errors I need help with and I'm not sure that an event emitter is the correct abstraction but it works. There's probably a better library to use than the default `events`.

I'm open to using a different abstraction to propagate the updates.

The subscription stuff is heavily based on https://github.com/facebook/react/tree/master/packages/create-subscription


FezVrasta:
Thanks for the PR.

May we use just the React context to handle the subscription?

Then we can have the context hold a list of `this.state.popperInstance.scheduleUpdate` of each of the children that will get called on each Popper.js update.

We'll have to figure out a way to decide if the `Popper` needs to use the Context.Provider (top level) or Context.Consumer (nested).

How does it sounds?

tyscorp:
@FezVrasta how's that? I think subscribing instances is better because otherwise you'll have to unsubscribe/resubscribe every time the Popper.js instance is recreated.

FezVrasta:
Thanks. It looks much better! I'll get back to you tomorrow

src/Popper.js
```diff
@@ -152,25 +160,70 @@ export class Popper extends Component<PopperProps, PopperState> {
...
+      this.unsubscribe();
+      this.setState({
+        unsubscribe: this.props.subscribe(this),
+      });
+    }
+  }
+
+  onUpdate() {
+    this.state.subscribers.forEach(popper =>
+      popper.state.popperInstance.update()
```
FezVrasta: this should be `scheduleUpdate`

FezVrasta:
@tyscorp I updated your PR with some code cleanup.

We need some unit tests now.

src/Popper.js
tyscorp: When I used `scheduleUpdate` there was a delay. `update` fixed it.

tyscorp: When I used `scheduleUpdate` there was a delay. `update` fixed it.

src/Popper.js
```diff
@@ -154,22 +169,49 @@ export class Popper extends Component<PopperProps, PopperState> {
...
-      style: this.getPopperStyle(),
-      placement: this.getPopperPlacement(),
-      arrowProps: {
-        ref: this.setArrowNode,
-        style: this.getArrowStyle(),
-      },
-    });
+    return (
+      <PopperContext.Provider
+        value={{ subscribe: this.subscribe, unsubscribe: this.unsubscribe }}
```
tyscorp: This will change context every render :)

FezVrasta: Right! Please feel free to fix it, it should be enough to add a class property to the component with the two methods as properties:

FezVrasta:
Is there still interest in this feature or should we close it?

FezVrasta:
There are some tests already in the repository, you may use them as baseline. 

To test this I think you may need to manually trigger the update method of Popper.js and check that all the necessary `Popper` components get updated accordingly.

tyscorp:
I'm having trouble mocking the inner component. I can't do it on the prototype because it needs to be an instance method.

Here's what I have so far:
```js
it('propagates updates through children', () => {
  const referenceElement2 = document.createElement('div');
  const wrapper = mount(
    <InnerPopper referenceElement={referenceElement}>
      {({ ref, style, placement, arrowProps }) => (
        <div ref={ref} style={style} data-placement={placement}>
          <div {...arrowProps} />
          <Popper referenceElement={referenceElement2}>
            {({ ref, style, placement, arrowProps }) => (
              <div ref={ref} style={style} data-placement={placement}>
                <div {...arrowProps} />
              </div>
            )}
          </Popper>
        </div>
      )}
    </InnerPopper>
  );
  // somehow mock inner Popper's onUpdate
  // const onUpdate = jest.mock(???)

  const instance = wrapper.state('popperInstance');
  instance.scheduleUpdate();

  expect(onUpdate).toBeCalled();
});
```

FezVrasta:
I think you can simply `jest.spyOn` it instead?

tyscorp:
How do I get a reference to use with `jest.spyOn`?

FezVrasta:
I thought that `jest.spyOn(wrapper.instance(), 'onUpdate')` would do the trick?

tyscorp:
That is the reference to the `InnerPopper`, we need to spy on the nested popper instance.

FezVrasta:
Oh sorry I didn't get it, I'm not sure what's the best way to proceed. Maybe find a different way to check your assertion.

PR #138, opened 2018-04-11
docs: Updated demo

FezVrasta:
demo

I redesigned the demo page and added two examples that use the new API.

We can add more examples later.


TheSharpieOne:
✨

PR #144, opened 2018-04-19
Fix #143

### LINKED ISSUE #143: v0.x demo does not build

1.  `npm install`
2.  `npm run demo`
3.  Build error:
```
react-popper/demo/animated.jsx:68:2: /Users/nijk/dev/frontend/react-popper/demo/animated.jsx: Missing class properties transform.
  66 |
  67 | class AnimatedExample extends PureComponent {
> 68 |   state = {
     |   ^
  69 |     isOpen: false,
  70 |   }
  71 |
```

The demo should build and run (I'm using the latest major node version)

### What went wrong?
I believe that the `babel-plugin-transform-class-properties` is required

### Any other comments?
I've also noticed that `package.json` is not handled either in the repo or via `.gitignore`

### Packages versions

- Popper.js: v0.x HEAD
- react-popper: n/a

nijk:
See #143

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,4 @@
 {
-  "presets": [["env", { "modules": false }], "stage-2", "react"]
+  "presets": [["env", { "modules": false }], "stage-2", "react"],
+  "plugins": ["transform-class-properties"]
 }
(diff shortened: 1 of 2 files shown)


FezVrasta:
stage-2 already includes that plugin 🤷‍♂️

PR #146, opened 2018-04-21
fix: v0.x dist builds to ensure SSR works properly and popper.js is included in the dist builds

### LINKED ISSUE #145: v0.10.1 - SSR breaks because of React-Popper + No /dist builds in 1.0beta5
The UMD build is broken in v0.10.1 because of:

`Manager.propTypes = process.env.NODE_ENV !== "production" ? {`

The production build should use plugin that optimizes and removes this code ahead of time. In Webpack it was the Define plugin. Then the remove prop-types plugin will correctly wrap it and leave it as `{}`.

What's the best way to resolve this?

0.10.1
> Rollup. I can find the corresponding plugin to ensure it optimizes it out and submit a PR for an immediate patch fix. Or are you not releasing updates anymore?

1.x.x
>  What's the situation here? There's currently no `/dist`. This is a huge problem. Reactstrap utilizes external opt-in dependency for Bootstrap and is a huge community. https://github.com/reactstrap/reactstrap

Ref: #142

virgofx:
Fixes #145 

After 0.8.3, the new dist build using Rollup config failed to include popper.js and missed some optimization settings for ensuring SSR rendering works correctly.

Updated:
* SSR rendering works using UMD build.
* Ensure /dist builds contain `popper.js`
* Removal of `prop-types` doesn't use wrap mode which would leave an empty `{}`. Instead, completely removed saving production builds a few bytes. Note that because the use of old `contextTypes`, prop-types will still be required until new Context 16.3 is used. At that point, production builds could completely reduce the dependency for prop-types.

@FezVrasta Can I create a separate PR that re-includes these rollup changes in the master branch such that /dist builds can continue to work once v1 becomes stable. Currently, the dist build was removed in the latest and will prevent 3rd party libraries that want to include popper capability via `react-popper` statically. Additionally, since it looks like work is being done to update context, the dist dependency on prop-types will go away 100%. I can also update the readme to explain the important/how to use for 3rd parties referencing statically as we do in Reactstrap.

/cc @TheSharpieOne  Note pending successful merge + release, we need to update the [readme](https://github.com/reactstrap/reactstrap/blob/master/README.md#example) in the next release to point to `.../0.10.2/react-popper.umd.min.js` (Note the `umd` portion) and update the default codepen JS references examples to all have the latest as well.


rollup.config.js
```diff
@@ -1,20 +1,34 @@
...
-    process.env.MINIFY ? minify() : false,
+    replace({
+      'process.env.NODE_ENV': JSON.stringify('production')
+    }),
+    process.env.NODE_ENV === 'production' ? minify({
+      comments: false,
+    }) : false,
   ].filter(Boolean),
-  external: ['react', 'react-dom', 'prop-types', 'popper.js'],
+  external: ['react', 'react-dom', 'prop-types'],
```
FezVrasta: We specifically want to leave popper.js out the build so that people can use their own version, mock it and reduce the final bundle size.

Why do you think it should be bundled?

virgofx: Because that's how Reactstrap and various other 3rd party libraries are using `react-popper` in a static sense. Essentially, think about a lot of frameworks and applications that are using React ... they want to add drop-in tooltip support and not worry about having to include an underlying tooltip api separately. 

As a compromise, I wouldn't mind creating 2 versions. That's actually what we do in Reactstrap as well ... we have a 'full.js' version.  That way people could include statically without popper.js implementation or a full drop-in that just works. Let me know if you're okay with that and I'll update as well as update the readme.

virgofx:
Not sure how to fix the travis-ci Rake issues (they're non-related to my changes).

Updated `./dist/` directory output with latest changes:

```
 14K Apr 22 15:02 react-popper.cjs.js
 17K Apr 22 15:02 react-popper.cjs.js.map
7.4K Apr 22 15:03 react-popper.cjs.min.js
 15K Apr 22 15:03 react-popper.cjs.min.js.map
 96K Apr 22 15:02 react-popper.full.cjs.js
145K Apr 22 15:02 react-popper.full.cjs.js.map
 31K Apr 22 15:03 react-popper.full.cjs.min.js
130K Apr 22 15:03 react-popper.full.cjs.min.js.map
102K Apr 22 15:02 react-popper.full.umd.js
154K Apr 22 15:02 react-popper.full.umd.js.map
 27K Apr 22 15:03 react-popper.full.umd.min.js
129K Apr 22 15:03 react-popper.full.umd.min.js.map
 16K Apr 22 15:02 react-popper.umd.js
 19K Apr 22 15:02 react-popper.umd.js.map
6.9K Apr 22 15:03 react-popper.umd.min.js
 15K Apr 22 15:03 react-popper.umd.min.js.map
```

virgofx:
ping @FezVrasta Any chance you can review so we can get this merged with a new build?

virgofx:
@FezVrasta Any chance this can get merged and a new release cut?

FezVrasta:
I don't think I understand the reasoning behind the "full version", it still requires React, React-DOM and Prop-Types. What's the point of having a version tha bundles a single dependency and no all the others?

react-popper never meant to ship its own version of Popper.js, it was a misconfiguration, I'd really like to avoid to have issues open on each Popper.js release asking me to release a new "full version" with the updated Popper.js release...

virgofx:
I can do that ... it just makes it easier to drop-in one file instead of two. I only included it that way because it was a regression without understanding that was the original intent. Upstream libraries will now need to be clear to include 2 files (`popper.js`  + `react-popper`)

virgofx:
Updated. I moved the umd builds into `/dist/umd` for consistency with `popper.js`.

```
./dist
├── react-popper.js
├── react-popper.js.map
├── react-popper.min.js
├── react-popper.min.js.map
└── umd
    ├── react-popper.js
    ├── react-popper.js.map
    ├── react-popper.min.js
    └── react-popper.min.js.map
```

virgofx:
ping @FezVrasta

FezVrasta:
Popper.js is a very bad example of dist folder structure 😅

I'd say that, or we keep everything at top level, or we keep each module type in its own folder

virgofx:
You pick --- I was trying to be consistent. Additionally, keeping a module in each folder is best... That's how react does it:

https://unpkg.com/react@16.3.2/

FezVrasta:
Yeah sorry, I'd like to change Popper.js but it would be a breaking change and people would get pissed off :-/

Let's try to do better here.

virgofx:
```
./dist
├── cjs
│   ├── react-popper.js
│   ├── react-popper.js.map
│   ├── react-popper.min.js
│   └── react-popper.min.js.map
└── umd
    ├── react-popper.js
    ├── react-popper.js.map
    ├── react-popper.min.js
    └── react-popper.min.js.map
```

FezVrasta:
Yes please, thanks!

virgofx:
Fez can you release 0.10.2 in the meantime when you get a chance? Will update master branch with PR soon. Thanks

FezVrasta:
0.10.2 published

PR #148, opened 2018-04-23
fix placement type in PopperProps.children

JLarky:
type is `undefined  | PopperJS.Placement`, not `?PopperJS.Placement` since that doesn't mean anything in typescript and was probably copied from Flow

fixes error that I get in typescript compiler:

```
node_modules/react-popper/typings/react-popper.d.ts
(24,16): JSDoc types can only be used inside documentation comments.
```

### DIFF
--- a/typings/react-popper.d.ts
+++ b/typings/react-popper.d.ts
@@ -21,7 +21,7 @@ interface PopperProps {
   children: (props: ({
     ref: (ref: HTMLElement | null) => void,
     style: React.CSSProperties,
-    placement: ?PopperJS.Placement,
+    placement?: PopperJS.Placement,
     outOfBoundaries: boolean | null,
     scheduleUpdate: () => void,
     arrowProps: {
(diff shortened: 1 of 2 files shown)


FezVrasta:
Thanks for the PR, we do have a test to verify that the ts types are right. May you check why isn't it detecting the problem please?

JLarky:
@FezVrasta I decided to test it :)
image

so it has nothing to do with typescript version, just with config option that was suppressing this warning in your CI tests

JLarky:
And I guess just to be clear what this PR does:
image

JLarky:
@FezVrasta do you need anything else for this PR?

giladgray:
this PR fixes #152!
i also just opened #153 with a more weighty refactor to the types to make them more useful.

JLarky:
@giladgray I'm just waiting for any of them to be merged so I can start using popper inside my project :) but I guess one more PR is needed that would remove `skipLibCheck` that was underlying issue that made #152 possible

FezVrasta:
Thanks for looking into it, may you adjust the TS configuration so that it correctly detects this problem in the future?

PR #158, opened 2018-05-01
Simplify flow type distribution

TrySound:
In this diff I replaced flow-copy-source with much more simple and less
verbose way to distribute flow types.

I suggest to reuse flow types which already exist in `src` folder by
reexporting them.

This also requires to publish `src`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,16 +9,17 @@
   "module": "lib/es/index.js",
   "typings": "typings/react-popper.d.ts",
   "files": [
-    "/dist",
-    "/lib",
-    "/typings/react-popper.d.ts"
+    "dist",
+    "lib",
+    "src",
+    "typings/react-popper.d.ts"
   ],
   "scripts": {
     "build": "npm run build:clean && npm run build:es && npm run build:cjs && npm run build:flow",
     "build:clean": "rimraf dist/ && rimraf lib/",
     "build:es": "cross-env BABEL_ENV=es babel src --ignore '*.test.js,__mocks__' --out-dir lib/es",
     "build:cjs": "cross-env BABEL_ENV=cjs babel src --ignore '*.test.js,__mocks__' --out-dir lib/cjs",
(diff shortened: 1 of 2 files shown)


FezVrasta:
What's wrong with the current approach?

TrySound:
It requires unnecessary tool and produces a lot of unnecessary files, which can be reused instead of duplication.

FezVrasta:
Yes but with the current method we avoid to ship tests and other unused stuff to npm. Also, it's quite common to ship the `.flow.js` files near the `.js` files, your approach is way less common.

TrySound:
Also forgot to mention. Currently flow are placed in es folder where they are absolutely useless. Developers usually import from `react-popper` not `react-popper/lib/es`.

FezVrasta:
If you import from `react-popper` and your bundler supports ES modules it should read them from `lib/es`, isn't that the case?

TrySound:
@FezVrasta `.js.flow` files are not bundler. They are consumed by flow. Flow don't care about `module` field.

FezVrasta:
Ok so we can simply change flow-copy-source to copy to cjs instead of es, but I'd keep it.

TrySound:
My way to distribute flow types is like bundling, only one file is placed in dist. Also I'm gonna to propose cjs and esm bundles instead of bunch of babeled files.

FezVrasta:
I prefer the current approach, really.

About bundles, I don't think there's need to generate them, anyway nobody is going to be able to use them in the browser directly.

TrySound:
Take a look it's quite common.
https://github.com/renatorib/react-powerplug/blob/master/package.json#L15
https://github.com/souporserious/react-measure/blob/rewrite/package.json#L13
https://github.com/atlassian/react-beautiful-dnd/blob/master/package.json#L37

UMD bundle can be used directly.

FezVrasta:
In any case, that would be an additional dist target, the ones we have should stay as they are.

FezVrasta:
I fixed the flow-copy-source command in #154.

If you'd like to add an UMD build please open a new PR, thanks!

PR #161, opened 2018-05-01
fix: Ensure "external-helpers" only included in Rollup builds. Fixes #1595

### LINKED ISSUE #160: "ReferenceError: babelHelpers is not defined" in version 0.10.2 (not in 0.10.1)

1.  I'm using material-ui beta 44 which uses react-popper as a dependency (bundled with create-react-app typescript scripts)

It should compile.

### What went wrong?

I get the following error:

```
ReferenceError: babelHelpers is not defined
(anonymous function)
node_modules/react-popper/lib/Manager.js:5
  2 | import PropTypes from 'prop-types';
  3 | 
  4 | var Manager = function (_Component) {
> 5 |   babelHelpers.inherits(Manager, _Component);
  6 | 
  7 |   function Manager() {
  8 |     var _ref;
```

### Any other comments?

Installing 0.10.1 manually fixes the problem. I can provide more info if necessary although given that the new version was only just released, I was hoping it would be an obvious fix.

### Packages versions

- Popper.js: 
- react-popper: 0.10.2

virgofx:
Confirmed `./lib/Manager.js` version built matches v0.10.1 and `/dist/*` matches v0.10.2

Fixes #160 

/cc @TrySound

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -21,7 +21,13 @@ const baseConfig = (outputFormat) => {
     input: 'src/index.js',
     plugins: [
       nodeResolve(),
-      babel(),
+      babel({
+        plugins: [
+          // Ensure "external-helpers" is only included in rollup builds
+          // Issue: https://github.com/rollup/rollup/issues/1595
+          'external-helpers',
+        ],
+      }),
       replace({
         'process.env.NODE_ENV': JSON.stringify('production')
       }),
--- a/.babelrc
+++ b/.babelrc
@@ -5,7 +5,6 @@
     "react"
   ],
   "plugins": [
-    "external-helpers",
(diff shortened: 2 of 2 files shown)


rollup.config.js
```diff
@@ -21,7 +21,13 @@ const baseConfig = (outputFormat) => {
     input: 'src/index.js',
     plugins: [
       nodeResolve(),
-      babel(),
+      babel({
+        plugins: [
```
FezVrasta: Is this going to extend the existing babelrc plugins list?

virgofx: Yes, it extends so it uses .babelrc then adds accordingly.

PR #162, opened 2018-05-01
Replace process.env.NODE_ENV in umd bundles

TrySound:
Seems like I missed this case.


FezVrasta:
Do we use process.env in the 1.x codebase?

TrySound:
@FezVrasta I found it in `warning` and `prop-types` packages which are bundled in umd.

FezVrasta:
I see! Thanks

PR #177, opened 2018-05-25
Upgrade to babel7

TrySound:
The size is slightly increased because of better helpers.


FezVrasta:
I'd rather wait for it to get out of beta first.

TrySound:
I'd say it's more stable than babel6. They support it only with critical stuff and didn't considered my critical stuff.

PR #180, opened 2018-05-25
Use reexport statement in entry point

TrySound:


FezVrasta:
I used this more verbose method because with the rexport Babel transpiles things in a funny way and things break for some users.

FezVrasta:
Please, before you spend time on something, discuss it in a new issue, so that we can decide if and how we want to proceed. I don't like you to waste time for nothing.

TrySound:
Rollup outputs cjs just perfectly without doubtful hacks like this.

PR #188, opened 2018-06-20
added transform-class-properties to fix issue with parcel

juanmaguitar:
I know this is not a final solution but I think it solves this problem for now.


FezVrasta:
Released as v1.0.1

PR #193, opened 2018-07-02
Document limiting overflow prevention to viewport

frontsideair:
Limit overflow prevention to viewport if you use a top-level portal for avoiding unnecessary replacements and performance.

I tried to add this to readme, but is it possible to detect it automatically? If so, I'd like to take a stab at it.


FezVrasta:
I think it's better to leave this to the user, if anything like this is really needed it should be integrated in popper.js

PR #197, opened 2018-07-09
perf: reduce renders by moving values off of PopperState

jquense:
This was originally meant to try and make it harder to hit the infinite loop with refs, but ended up with a small perf improvement by reducing the need to rerender when `popper` and `arrow` nodes are set. We don't actually need to update any children until _after_ the node updates have made their way back into popper, at which point the modifier will update state and re-render with the new styles

### DIFF
--- a/src/Popper.js
+++ b/src/Popper.js
@@ -41,7 +41,6 @@ export type PopperProps = {
 };
 
 type PopperState = {
-  popperInstance: ?PopperJS$Instance,
   data: ?Data,
 };
 
@@ -64,9 +63,10 @@ export class InnerPopper extends React.Component<PopperProps, PopperState> {
   };
 
   state = {
-    popperInstance: undefined,
     data: undefined,
   };
+
+  popperInstance: ?PopperJS$Instance;
 
   popperNode: ?HTMLElement = null;
   arrowNode: ?HTMLElement = null;
@@ -130,9 +130,8 @@ export class InnerPopper extends React.Component<PopperProps, PopperState> {
     this.state.data ? this.state.data.hide : undefined;
 
   initPopperInstance = () => {
-    const { popperNode} = this
+    const { popperNode, popperInstance } = this
     const { referenceElement } = this.props;
-    const { popperInstance } = this.state;
 
     if (referenceElement && popperNode && !popperInstance) {
       const popperInstance = new PopperJS(
@@ -140,28 +139,31 @@ export class InnerPopper extends React.Component<PopperProps, PopperState> {
         popperNode,
         this.getOptions()
       );
-      this.setState({ popperInstance });
+      this.popperInstance = popperInstance;
       return true;
     }
     return false;
   };
 
-  destroyPopperInstance = (callback: () => boolean) => {
-    if (this.state.popperInstance) {
-      this.state.popperInstance.destroy();
-    }
-    this.setState({ popperInstance: undefined }, callback);
+  destroy = () => {
+    if (!this.popperInstance) return false
+    this.popperInstance.destroy();
+    this.popperInstance = null
+    return true;
+  }
+
+  destroyPopperInstance = (callback: () => void) => {
+    if (this.destroy()) this.forceUpdate(callback)
+
   };
 
   updatePopperInstance = () => {
-    if (this.state.popperInstance) {
-      this.destroyPopperInstance(() => this.initPopperInstance());
-    }
+    if (this.destroy()) this.initPopperInstance();
   };
 
   scheduleUpdate = () => {
-    if (this.state.popperInstance) {
-      this.state.popperInstance.scheduleUpdate();
+    if (this.popperInstance) {
+      this.popperInstance.scheduleUpdate();
     }
   };
 
@@ -184,9 +186,7 @@ export class InnerPopper extends React.Component<PopperProps, PopperState> {
   }
 
   componentWillUnmount() {
-    if (this.state.popperInstance) {
-      this.state.popperInstance.destroy();
-    }
+    this.destroy()
   }
 
   render() {
--- a/src/Popper.js
+++ b/src/Popper.js
@@ -32,7 +32,6 @@ export type PopperChildren = PopperChildrenProps => React.Node;
 
 export type PopperProps = {
   children: PopperChildren,
-  init?: boolean,
   eventsEnabled?: boolean,
   innerRef?: getRefFn,
   modifiers?: Modifiers,
@@ -57,7 +56,6 @@ const initialArrowStyle = {};
 
 export class InnerPopper extends React.Component<PopperProps, PopperState> {
   static defaultProps = {
-    init: true,
     placement: 'bottom',
     eventsEnabled: true,
     referenceElement: undefined,
@@ -132,24 +130,20 @@ export class InnerPopper extends React.Component<PopperProps, PopperState> {
   getOutOfBoundariesState = () =>
     this.state.data ? this.state.data.hide : undefined;
 
-  destroy = () => {
-    if (!this.popperInstance) return false;
+  destroyPopperInstance = () => {
+    if (!this.popperInstance) return;
+
     this.popperInstance.destroy();
     this.popperInstance = null;
-    return true;
-  };
-
-  destroyPopperInstance = (callback: () => void) => {
-    if (this.destroy()) this.forceUpdate(callback);
   };
 
   updatePopperInstance = () => {
-    this.destroy();
+    this.destroyPopperInstance();
 
     const { popperNode } = this;
-    const { referenceElement, init } = this.props;
+    const { referenceElement } = this.props;
 
-    if (!referenceElement || !popperNode || !init) return;
+    if (!referenceElement || !popperNode) return;
 
     this.popperInstance = new PopperJS(
       referenceElement,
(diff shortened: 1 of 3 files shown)


jquense:
related to #185

jquense:
still a work in progress!

jquense:
ok I think this is all the low hanging fruit, I reorganized the popper init logic a bit, but some of it was unnecessary it just sort of happened as i was trying to understand the flow.

src/Popper.js
```diff
@@ -59,24 +57,37 @@ const initialArrowStyle = {};
 
 export class InnerPopper extends React.Component<PopperProps, PopperState> {
   static defaultProps = {
+    init: true,
```
jquense: I've added this so implementors can lazy initialize popper, such as in the case of tooltip or dropdown, only when the popper element is visible

FezVrasta: May you help me to understand better? If I set this boolean to `true` and then to `false, what happens?

FezVrasta:
Thank you for the PR, unfortunately these weeks I'm not at home so I have little time to look at my OSS projects, I'll try to get it merged as soon as possible, I just need to understand few things.

src/Popper.js
```diff
@@ -114,71 +125,60 @@ export class InnerPopper extends React.Component<PopperProps, PopperState> {
...
+    this.popperInstance = null;
+    return true;
   };
 
-  destroyPopperInstance = (callback: () => boolean) => {
-    if (this.state.popperInstance) {
-      this.state.popperInstance.destroy();
-    }
-    this.setState({ popperInstance: undefined }, callback);
+  destroyPopperInstance = (callback: () => void) => {
```
FezVrasta: I'm a bit confused by the `destroy` vs `destroyPopperInstance` methods, the latter also forces a React re-render, but I don't understand how do you decide when to use one rather than the other. (Also, we could give them more explicit names maybe)

src/Popper.js
FezVrasta: Also, is there any reason to prefer this rather than not making React render the whole component completely?

I mean:

```jsx
init = false;
[...]
{init && <Tooltip />}
```

rather than:
```jsx
<Tooltip init={false} />
```

jquense: Yeah I'm not set on this API yet, I'm still testing it out in terms of what we need for react-bootstrap.

Generally yes conditionally rendering the popper element would handle this case and honestly I think that's the best approach. I've added it to enable the bootstrap behavior of keeping drop-down menus rendered even when closed (initially). We could render the menu without the popper element intially but that would cause the drop-down to unmount and remount can update when it opens, since it's parent is changing

The other thing I wanted to enable is clearer control on when popper is initiated vs updated. I don't really think this API solves that well but it was my first attempt, and maybe schedule update is enough.

Let me play around with this a bit more

src/Popper.js
jquense: The reason for both is this. When destorying to update the popper instance, e.g. create a new one, we don't want to trigger a component update, the only thing the child cares about is if the position in styles change, and that will be triggered when the new instance hits the updateStateModifier

I think I left the other version, with the force update to not break the original component API, in case they were public methods. I can make them more explicit and remove if not needed 👍

jquense:
Hey no worries on getting this merge for my sake, I've got my own temporary fork published, react-bootstrap is using as we add v4 support. Review at your leisure, no stress.

giladgray:
this is hot!! 🔥 🔥

src/Popper.js
jquense: wrote the above on my phone, but here is an example for why a toggle prop may be preferable

```js
{show ? <Popper>{() => <DropdownMenu />}</Popper> : <DropdownMenu />}
```

if it is required that the menu always be rendered the above works but causes a remount on each `show` toggle vs staying mounted but updating.

FezVrasta: I guess what I'm trying to understand is in which case you want to switch from a positioned to not-positioned version of the same dropdown. I understand the Bootstrap case, but there's not dynamic switching involved there, there are some places where Popper.js is used and places where it's not, that's it, right?

src/Popper.js
FezVrasta: Thanks I'll play with it on Monday once I hit the office.

src/Popper.js
jquense: I don't usually want to do switch between positioned and not, I _do_ want to defer initializing popper until a user opens a tooltip or dropdown, so all the popper init work isn't done when the page loads, but no tooltips or dropdowns are visible.

The actual common case here is something like this: 

```js
{hasShownOnce ? <Popper>{() => <DropdownMenu />}</Popper> : <DropdownMenu />}
```

so once the dropdown is shown it, popper stays, but not until the first `show`. I will say tho that vanilla bootstrap does recreate `popper` on each show, i'm not sure why they do that vs doing a scheduleUpdate tho.

jquense: It may also be worth not running updates initiated from `onScroll` or window resize when the element isn't visible? Maybe popper already optimizes this?

FezVrasta: Umh okay... Last question (I hope), why do you need to render the dropdown even when you don't want to show it? Couldn't it be:

FezVrasta: Also, now that I think about it, Popper.js accepts a `eventsEnabled` option which may be used in place of this custom implementation maybe?

FezVrasta: In any case, maybe it would make sense to move this to a different PR and leave this one focused on perf improvement so that it's easier to review it, if it's a problem I can do it myself.

jquense: > Couldn't it be:
` {hasShownOnce && <Popper>{() => <DropdownMenu />}</Popper>}`

They only reason i'm a bit concerned with doing that, is that it's different from how bootstrap and react-bootstrap handle dropdowns by default. Given how many users RB has, i'm not confident that not mounting the DropdownMenu initially won't break someone's workflow. People put a ton of logic in their components and it's not clear that they won't need that to run immediately (maybe the dropdown item needs to attach global key handlers for keycode shortcuts i don't know).

I'm happy to try the `eventsEnabled` as an initial strategy, it's easy to add this API in the future if it's needed, and harder to remove it if it's not

shepmaster:
> make it harder to hit the infinite loop with refs

Looking forward to this improvement ❤️ !

shepmaster:
I've pulled this branch locally and it doesn't solve my case. However, I think you just want to apply yhr same type of transformation to `Manager`:

```diff
diff --git a/src/Manager.js b/src/Manager.js
index e6b3cfb..23e90d5 100644
--- a/src/Manager.js
+++ b/src/Manager.js
@@ -2,40 +2,30 @@
 import * as React from "react";
 import createContext, { type Context } from "create-react-context";

-export const ManagerContext: Context<{
-  getReferenceRef?: (?HTMLElement) => void,
-  referenceNode?: ?HTMLElement,
-}> = createContext({ getReferenceRef: undefined, referenceNode: undefined });
+type ManagerContextX = {
+    getReferenceRef?: (?HTMLElement) => void,
+    referenceNode?: ?HTMLElement,
+};
+
+export const ManagerContext: Context<ManagerContextX> =
+  createContext({ getReferenceRef: undefined, referenceNode: undefined });

 export type ManagerProps = {
   children: React.Node,
 };
-type ManagerState = {
-  context: {
-    getReferenceRef?: (?HTMLElement) => void,
-    referenceNode?: ?HTMLElement,
-  },
-};

-export default class Manager extends React.Component<ManagerProps, ManagerState> {
-  constructor() {
-    super();
-    this.state = {
-      context: {
-        getReferenceRef: this.getReferenceRef,
-        referenceNode: undefined,
-      },
-    };
-  }
+export default class Manager extends React.Component<ManagerProps> {
+  getContext = (): ManagerContextX => ({
+    getReferenceRef: this.getReferenceRef,
+    referenceNode: this.referenceNode,
+  })

   getReferenceRef = (referenceNode: ?HTMLElement) =>
-    this.setState(({ context }) => ({
-      context: { ...context, referenceNode },
-    }));
+    this.referenceNode = referenceNode;

   render() {
     return (
-      <ManagerContext.Provider value={this.state.context}>
+        <ManagerContext.Provider value={this.getContext()}>
         {this.props.children}
       </ManagerContext.Provider>
     );
```

shepmaster:
May I ask why the refs were put into state in the first place? From [my little knowledge](https://reactjs.org/docs/refs-and-the-dom.html), refs have always been directly stored as an instance property. Obviously, there's a good reason to put them into the state or else this library wouldn't have done so, but I don't know why. Can anyone point me in the right direction to learn more?

jquense:
The refs really need to be in state, or more directly, trigger an update, so the context consumers down the tree are updated. Otherwise they may not see that reference element was set. The case is a bit different for Popper where the refs aren't passed down anywhere

jquense:
Ok updated without the `init` option for now. should be good 👍

FezVrasta:
@jquense I merged your PR but I see this behavior as soon I switch placement in the demo page:

image

May you take a look if you have time?

jquense:
yup i'll check it out 👍

seancurtis:
Any idea when this will be released? The re-rendering caused by the reference setting is blocking me. Would a 1.0.1 be feasible?

FezVrasta:
I have to merge another PR that fixes a regression first, I need to dedicate some time to it.

FezVrasta:
No it's really broken right now

jquense:
If you really need it now, `@react-bootstap/react-popper` is a bleeding edge version with all my PRs

seancurtis:
@FezVrasta is there anything I can do to help get master in a good state?

jquense:
Prs are already up! It's always helpful to have some additional eyes on it and testing if you wanna pull that and try

seancurtis:
@jquense where is the src for `@react-bootstrap/react-popper`?

FezVrasta:
Released as v1.0.1

PR #207, opened 2018-08-09
fix: position update

FezVrasta:


FezVrasta:
I think the new `placement` state is enough to trigger an updated when needed, that's why it's working without the `getDerivedStateFromProps`

FezVrasta:
Released as v1.0.1

PR #211, opened 2018-08-22
Always recreate popper instance when popperNode changes

wyozi:
In our codebase there's a subset of tooltips that do not work properly. Their tooltips work correctly on the first hover but on subsequent hovers only appear after you move your mouse over the element and then away from it.

After some debugging I traced the problem to the `!this.popperNode` condition inside `getPopperStyle()`; replacing it with `false` made the tooltip appear properly on all hovers. That does not seem like a proper fix, however, so I figured updating popper instance every time after popperNode updates would work, and that it does. 

As far as I understand this change results in correct behavior, since Popper instance stores a reference to the popperNode and if the node updates, so should the instance?

### DIFF
--- a/src/Popper.js
+++ b/src/Popper.js
@@ -79,7 +79,7 @@ export class InnerPopper extends React.Component<PopperProps, PopperState> {
     safeInvoke(this.props.innerRef, popperNode);
     this.popperNode = popperNode;
 
-    if (!this.popperInstance) this.updatePopperInstance();
+    this.updatePopperInstance();
   };
 
   setArrowNode = (arrowNode: ?HTMLElement) => {


src/Popper.js
```diff
@@ -79,7 +79,7 @@ export class InnerPopper extends React.Component<PopperProps, PopperState> {
     safeInvoke(this.props.innerRef, popperNode);
     this.popperNode = popperNode;
 
-    if (!this.popperInstance) this.updatePopperInstance();
+    this.updatePopperInstance();
```
FezVrasta: What's the point of updating a non existing popper instance?

wyozi: So there should rather be a check for if `popperNode` is not defined than?

wyozi: Sorry, misunderstood. Doesn't that line of code originally mean that the popper instance is only updated if it's non existing?

bodyflex: @FezVrasta before this fix, on L80 the `popperNode` gets updated on the class, but not inside the `popperInstance` which leaves the class in an inconsistent state. when the `popperNode` changes, we need to reflect that in the `popperInstance` as well, so it's necessary to call `updatePopperInstance()` (which is actually a misleading name since it updates the instance).

right now updating `popperNode` does not work when the popperInstance has already been created.

cy6erskunk:
@FezVrasta it would be awesome if you could publish a bugfix version containing this fix!

PR #219, opened 2018-09-25
Expose "update" method from PopperJS

quentez:
Right now only the `scheduleUpdate` method is exposed.
This makes it difficult to build frame-perfect behaviors that rely on Popper updating.


FezVrasta:
Thanks for the PR, may you attach a codepen or working demo of a case where you need to call `update` rather than `scheduleUpdate`?

PR #224, opened 2018-10-09
export defaults from popperJS

asturur:
Why?

in my app i need to look at popperJS default values in order to pass down props to <Popper> component.

To do so, i have to add a reference to popperJS library in my project that pulls in a copy of popperJS other than the one that is imported from react-popper ( probably because it is not the exact same version )

Since <Popper> takes a prop of type modifiers, and those modifiers are a complex object stored in popperJS.Defaults, is comfortable to have the default value for that object at hand.

Unless i m missing a better way to get them from react-popper

### DIFF
--- a/src/Popper.js
+++ b/src/Popper.js
@@ -209,8 +209,8 @@ export class InnerPopper extends React.Component<PopperProps, PopperState> {
   }
 }
 
-const placements = PopperJS.placements;
-export { placements };
+const { placements, Defaults } = PopperJS;
+export { placements, Defaults };
 
 export default function Popper(props: PopperProps) {
   return (
--- a/src/Popper.js
+++ b/src/Popper.js
@@ -209,8 +209,8 @@ export class InnerPopper extends React.Component<PopperProps, PopperState> {
   }
 }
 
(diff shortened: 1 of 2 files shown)


asturur:
Ps: i have no understanding of how i m supposed to update the tests. I tried but i could not find where the statics are.

asturur:
Maybe exporting a full reference to the popperJs lib would be better.

asturur:
I changed it to full PopperJS export

FezVrasta:
I don't think it is needed, you shouldn't rely on transitive dependencies in your code. If you need Popper.js, add Popper.js to your dependencies. Your package manager will make sure to use a single version.

asturur:
well it will and it won't.
Depends on what your package lock says and how you update them.
I do not need popper.js but the defaults object structure of the props is not accessible and hard to guess, and not really right to write it all again.

So if i have to pass the modifiers as a prop, and i want to change just a value, i take the defaults, clone it, modify my value and pass as a prop.

Where do i take the defaults from?

FezVrasta:
as I said, you should just add popper.js to your dependencies, that's the solution I suggest to follow.

PR #231, opened 2018-10-18
More explicit referenceElement setting on Popper.js

gwyneplaine:
Hello, first and foremost thank you so much for this library! 
This is a small non-intrusive change to how prop.referenceElement and referenceNode passed through context currently interact within the Popper.js file. 

For context, within atlassian we currently have a component that's a thin wrapper around react-popper that enforces certain internal configuration opinions. This wrapper supports both using the Reference component to instantiate a referenceNode in context; as well as the referenceElement prop of the Popper component. At the moment declaring a default value of undefined on the referenceElement prop overrides the referenceNode that may be available through context. This seems like an unintended side effect of prop spreading. 

This is a small change to ensure that if referenceElement is not defined, we always fallback to the referenceNode in context. Added two quick tests to assert this. 

Thanks again for the lib!

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,25 +1,25 @@
 {
   "dist/index.umd.js": {
-    "bundled": 49411,
-    "minified": 17642,
-    "gzipped": 5583
+    "bundled": 49703,
+    "minified": 17814,
+    "gzipped": 5717
   },
   "dist/index.umd.min.js": {
-    "bundled": 24106,
-    "minified": 9863,
-    "gzipped": 3275
+    "bundled": 25362,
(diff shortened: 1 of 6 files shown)


FezVrasta:
Thank you for the PR and sorry for the delay.

May you document this change in the readme please?

gwyneplaine:
Hi @FezVrasta, thanks for the reply! I'm unsure of what would be useful to document in the readme here, as this is a bugfix, and ultimately doesn't affect the existing API. Happy to add to the readme if you still feel its necessary; let me know what would be useful to document 👍

FezVrasta:
Umh you're right sorry

Issue #239, opened 2018-11-06
remove @types/react from dependencies


1pete:
From #234, I think library shouldn't include types of other dependencies. If types has to be in dependencies, it has to be in react project itself, so other dependencies those depend on it, don't have to make the redundancy.

I have tons of examples
react-dropzone: https://github.com/react-dropzone/react-dropzone/blob/master/package.json
react-datepicker: https://github.com/Hacker0x01/react-datepicker/blob/master/package.json
reactstrap: https://github.com/reactstrap/reactstrap/blob/master/package.json

But if it's `@type/react-popper`, then yeah, it'd make a bit sense.

FezVrasta:
react-popper types require React types, so it's correct to have them in the dependencies list.

Strnadj:
@FezVrasta what is the solution? I always have to manually remove `@types/react` from `node_modules/react-popper` otherwise it conflicts with `@types/react` installed in project dependencies. It causes very ugly errors like:

```
ERROR in [at-loader] ./node_modules/@types/react/index.d.ts:2305:19 
    TS2320: Interface 'Element' cannot simultaneously extend types 'ReactElement<any>' and 'ReactElement<any>'.
  Named property 'type' of types 'ReactElement<any>' and 'ReactElement<any>' are not identical.
```

as @1pete mentions, it can be solved (as examples). 

Next thing is types (`@types/react`) are in `dependencies` and `devDependencies` ... twice

Strnadj:
What about moving TypeScript definitions to https://github.com/DefinitelyTyped/DefinitelyTyped and that is all... For those who want it, install `@types/react-popper`....

FezVrasta:
Yes that'd be a good idea but then I will not even try to maintain them anymore.

Honestly I thought it would be easy to ship those types with the library, with Popper.js (which is not typed) I just added a `popper.js.flow` file with the types and it just works, I thought it would have been the same with TS here...

rogierschouten:
@FezVrasta  the real solution is to make  @types/react both a peerDependency and a devDependency.  The peerDependency should preferably have a "*" version or a version range.

1pete:
@rogierschouten no, that is an absurd move because every application has to declare @types/react as dependencies regardless it uses typescript or not

I have 2 suggestions
1. **(Recommended) In this project, just declare @types/react in devDependencies** (or **completely remove from everywhere**)
  so applications those use typescript and this library, they has to add @types/react to their own package.json
[...]

PR #242, opened 2018-11-15
Fix for #239

rogierschouten:
tried & tested locally

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,10 +55,10 @@
     "popover"
   ],
   "peerDependencies": {
+    "@types/react": "*",
     "react": "0.14.x || ^15.0.0 || ^16.0.0"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


FezVrasta:
Thanks for the PR!

I'm sorry but I'm not going to merge any change related to TypeScript without an extensive test suite attached to the PR. This is required because since I added TS to this project, every single time a PR that edited TS has been merged something broke.

rogierschouten:
Hmm... any idea as to what that test should look like?  I guess you'd want to see that a fake new NPM module would compile against a published version of react-popper - otherwise it wouldn't say much

rogierschouten:
@FezVrasta good point. Sorry for causing you extra work. Maybe indeed move typings to DefinitelyTyped

PR #244, opened 2018-11-16
refactor: getReferenceRef -> setReferenceNode to avoid confusion

pd4d10:
`getReferenceRef` actually set the reference node and return nothing, rename it to `setReferenceNode` to avoid confusion


FezVrasta:
Sorry for the delay and thanks for the PR!

Just to be sure, this doesn't change any external/user-facing property right?

pd4d10:
Yes, it is a private method

PR #245, opened 2018-11-18
build: remove @types/react from dependencies

### LINKED ISSUE #239: remove @types/react from dependencies

pd4d10:
fix #239

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,6 @@
     "react": "0.14.x || ^15.0.0 || ^16.0.0"
   },
   "dependencies": {
-    "@types/react": "^16.1.0",
     "babel-runtime": "6.x.x",
     "create-react-context": "^0.2.1",
     "popper.js": "^1.14.4",


mykas:
Can we merge this? we have issues with this too.

mykas:
Or at least put it in devDependencies.

piecyk:
@FezVrasta This is correct solution, more at https://github.com/Microsoft/TypeScript/issues/9731#issuecomment-232864820 
Let's merge this ?

FezVrasta:
I'm quite sure this will break something else, but let's do that...

piecyk:
Eee, have more fait 🤗

PR #251, opened 2018-12-04
chore: remove dependency on create-react-context and babel-runtime

MarshallOfSound:
This PR upgrades a bunch of dependencies (mostly babel and react) to facilitate the removal of `create-react-context` and `babel-runtime`.

This changes bumps the react peer dependency requirement to `react^16.3.0` in order to guarantee built in support for `React.createContext`.  This lowers bundle size and gives us the benefits of using a React built-in.

It also bumps us to babel 7 so we can drop `babel-runtime` and use `@babel/runtime` instead so folks on babel 7 (most people at this point) don't end up with 2 babel runtimes in their dist which can cause issues.


src/Manager.js
```diff
@@ -1,11 +1,7 @@
 // @flow
 import * as React from "react";
-import createContext, { type Context } from "create-react-context";
 
-export const ManagerContext: Context<{
-  getReferenceRef?: (?HTMLElement) => void,
-  referenceNode?: ?HTMLElement,
-}> = createContext({ getReferenceRef: undefined, referenceNode: undefined });
+export const ManagerContext = React.createContext<*>({ getReferenceRef: undefined, referenceNode: undefined });
```
FezVrasta: The existential type (*) is deprecated, may you define a proper type here?

piecyk:
IMHO it would be better to split this PR, one about removing `create-react-context` and other about updating dependencies.

Was thinking about how to describe polyfill the createContext API but hard to do that in nice way 🤔  One hacky option is to relay on that `React` is exported as `object` so we can update it 🙈 

https://codesandbox.io/s/5k6p25lxqk

FezVrasta:
Maybe it's just better to leave the dependency in place, in any case it's not going to add any weight at runtime with dead code elimination.

Once we move to the new hooks based API I was thinking about we can cut the cord with React 15

piecyk:
Yeah, maybe let's leave it for now... But we can improve it a bit, if we change to 
```js
// @flow
import createReactContext, { type Context  } from "create-react-context";

const createContext = (React: any).createContext || createReactContext

export const ManagerContext: Context<{
  getReferenceRef?: (?HTMLElement) => void,
  referenceNode?: ?HTMLElement,
}> = createContext({ getReferenceRef: undefined, referenceNode: undefined });
```
then we will in control of the `create-react-context`, then user can override it at build phase with void
or event more suppress it in yarn with `resolutions` replacing it zero dependency for example  

```
  "resolutions": {
    "create-react-context": "https://github.com/piecyk/empty-module.git#0.0.1"
  }
```

What do you think ?

piecyk:
Also it looks like we have two open PR's regarding updating dependencies, babel7 🎉 @MarshallOfSound can you revert the changes for create-react-context and we can release it sooner?

FezVrasta:
I merged the other PR that upgraded Babel to 7.

I think it makes sense to close this one and ask for one just for the babel-runtime removal (if it doesn't impacts too negatively on the bundle size)

Sorry for the confusion!

PR #253, opened 2018-12-08
Fix `setPopperNode` only to update popper instance when for HTMLElement

### LINKED ISSUE #250: Infinite render loop with Downshift
### Reproduction demo

https://codesandbox.io/s/wzn6ppylml

1.  Open the reproduction above
2.  *OPEN DEVTOOLS*, or the tab will hang
3.  Start typing into the input

Popper element appears and doesn't cause any trouble.

### What went wrong?

There's an infinite state update loop which seems related to the `ref` from Popper being combined with the internal `ref` from Downshift.

As you can see, I'm using Downshift's recommended pattern of passing all props (including `ref`) through their `getMenuProps()` helper function.

This runs Popper's `ref` through their [combiner function](https://github.com/paypal/downshift/blob/master/src/utils.js#L108) ([caller](https://github.com/paypal/downshift/blob/master/src/downshift.js#L823)) and spits out a new, composed function to the final `ref` prop.

As soon as that `ref` is called, Popper begins an infinite loop. The key problem is [this](https://github.com/FezVrasta/react-popper/blob/master/src/Popper.js#L77) guard. We run into the infamous functional ref behavior where the same ref is [called twice with `null`, then the element](https://github.com/facebook/react/issues/4533). Since it's the same function, this forces Popper to oscillate between `null` and the correct element reference and constantly update its state.

### Any other comments?

We have been using this library for a while and only encountered this problem recently. I suspect this was because of [this change](https://github.com/FezVrasta/react-popper/commit/48197a7b261f220af463086fd1b6b30acc9e6169#diff-57c57db49f6c673250f466945a91cf07) which may have removed a guard that stopped the loop early.

Also, apologies for the terse tone here. It's been a long day of trying to diagnose what's going on here. I mean no ill will toward anyone and would be happy to dig in a bit into the root of this behavior and possible mitigation. Cheers for having very readable source code.

### Why am I opening this against Popper and not Downshift?

This problem isn't isolated to Downshift specifically, but is a product of how react-popper works and some assumptions it makes. Downshift happens to encourage patterns which break it.

That said, I'm going to try some experiments with memoization and see if I can find a good mitigation strategy. If it turns out that memoizing `getMenuProps` is reasonable and could help solve the problem, I may float that by the Downshift maintainers.

I'd be happy enough if you decided to close this as wontfix, being satisfied that I documented this behavior so that hopefully future searchers might find some specifics (including, hopefully, some workarounds which I will comment here if I find them).

I still think Popper could document this gotcha more explicitly though, to aid others who run across the issue in different contexts. I've seen other issues here along this same line (#199, etc), but if I didn't know the problem I was looking for, I would not have realized they were related.

Perhaps there's also a change the library could make to prevent this happening in the first place (like checking for a `null` popperNode and not updating state).

### Packages versions

- Popper.js: whatever's bundled
- react-popper: 1.3.0

piecyk:
closes #250 

@FezVrasta have a look, need to tweak bit the popper.js mock. We want to call `updateStateModifier` because it was triggering the infinite loop with setState, after the instance was recreated... Also updated few types to be more strict from `?HTMLElement` to `HTMLElement | null` as the void type is not quite correct there.

### DIFF
--- a/src/Popper.js
+++ b/src/Popper.js
@@ -11,7 +11,7 @@ import type { Style } from 'typed-styles';
 import { ManagerContext } from './Manager';
 import { safeInvoke, unwrapArray } from './utils';
 
-type getRefFn = (?HTMLElement) => void;
+type getRefFn = (HTMLElement | null) => void;
 type ReferenceElement = ReferenceObject | HTMLElement | null;
 type StyleOffsets = { top: number, left: number };
 type StylePosition = { position: 'absolute' | 'fixed' };
@@ -59,7 +59,7 @@ export class InnerPopper extends React.Component<PopperProps, PopperState> {
   static defaultProps = {
     placement: 'bottom',
     eventsEnabled: true,
-    referenceElement: undefined,
+    referenceElement: null,
     positionFixed: false,
   };
 
@@ -68,13 +68,13 @@ export class InnerPopper extends React.Component<PopperProps, PopperState> {
     placement: undefined,
   };
 
-  popperInstance: ?Instance;
+  popperInstance: Instance | null;
 
-  popperNode: ?HTMLElement = null;
-  arrowNode: ?HTMLElement = null;
+  popperNode: HTMLElement | null = null;
+  arrowNode: HTMLElement | null = null;
 
-  setPopperNode = (popperNode: ?HTMLElement) => {
-    if (this.popperNode === popperNode) return;
+  setPopperNode = (popperNode: HTMLElement | null) => {
+    if (!popperNode || this.popperNode === popperNode) return;
 
     safeInvoke(this.props.innerRef, popperNode);
     this.popperNode = popperNode;
@@ -82,8 +82,8 @@ export class InnerPopper extends React.Component<PopperProps, PopperState> {
     this.updatePopperInstance();
   };
 
-  setArrowNode = (arrowNode: ?HTMLElement) => {
-    if (this.arrowNode === arrowNode) return;
+  setArrowNode = (arrowNode: HTMLElement | null) => {
+    if (!arrowNode || this.arrowNode === arrowNode) return;
     this.arrowNode = arrowNode;
 
     if (!this.popperInstance) this.updatePopperInstance();
@@ -94,10 +94,7 @@ export class InnerPopper extends React.Component<PopperProps, PopperState> {
     order: 900,
     fn: (data: Object) => {
       const { placement } = data;
-      this.setState(
-        { data, placement },
-        placement !== this.state.placement ? this.scheduleUpdate : undefined
-      );
+      this.setState({ data, placement });
       return data;
     },
   };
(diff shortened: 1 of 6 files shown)


src/Popper.js
```diff
@@ -68,22 +68,22 @@ export class InnerPopper extends React.Component<PopperProps, PopperState> {
...
     this.updatePopperInstance();
   };
 
-  setArrowNode = (arrowNode: ?HTMLElement) => {
-    if (this.arrowNode === arrowNode) return;
+  setArrowNode = (arrowNode: HTMLElement | null) => {
+    if (!arrowNode || this.arrowNode === arrowNode) return;
     this.arrowNode = arrowNode;
 
     if (!this.popperInstance) this.updatePopperInstance();
```
piecyk: Don't quite understand way we have this code, in what situation we will set the arrow and we don't have the `popperInstance` ?  IMHO would simplify it and just use 

```js
  setArrowNode = (arrowNode: HTMLElement | null) => {
    this.arrowNode = arrowNode;
  };
```

src/Popper.js
```diff
@@ -59,7 +59,7 @@ export class InnerPopper extends React.Component<PopperProps, PopperState> {
   static defaultProps = {
     placement: 'bottom',
     eventsEnabled: true,
-    referenceElement: undefined,
```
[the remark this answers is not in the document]
piecyk: To be the same value as it can be set by `ref` setter in React, there we have `(HTMLElement | null) => void;` it can't be undefined. Type in props is `referenceElement?: ReferenceElement,` so if the `referenceElement` is void it will be type `ReferenceElement` so `ReferenceObject | HTMLElement | null` that's way changed it to `null`

src/Popper.test.js
```diff
@@ -71,7 +71,7 @@ describe('Popper component', () => {
     const referenceElement = document.createElement('div');
     expect(() =>
       mount(
-        <InnerPopper referenceElemen={referenceElement}>
+        <InnerPopper referenceElement={referenceElement}>
```
piecyk: Thanks to `typo` test was passing. Not a `flow` user... but typescript would throw an error `Property 'referenceElemen' does not exist on type ...`

FezVrasta: referenceElem is not disallowed by the component. If TS thrown that kind of error it's wrong

src/Popper.js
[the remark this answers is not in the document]
FezVrasta: In flow `?Type` can be Type, null or void. Please keep it as it was before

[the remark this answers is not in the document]
piecyk: Yep i know, as mention in most cases the values will be never void that's way this more explicit change. Will revert 👍

src/Popper.test.js
piecyk: We have `referenceElement` on PopperProps not `referenceElemen` as it looks like typo, just was wondering way flow don't warn about it... but not related to this PR

FezVrasta: Yes I noticed it. It doesn't warn because the typoed prop is not disallowed explicitly

piecyk:
When looking closer found out that we have some not needed renders in Popper when placement is not changed. Small change https://github.com/piecyk/react-popper/commit/9d44f5192b7102a6e0296c74838826cbd0d0889c I know the `test` for it is not very nice 🤔 Basic `scheduleUpdate` update was called two times after placement was changed... I can create issue for this or submit a PR

piecyk:
@FezVrasta What do you think about this ?

FezVrasta:
It looks good to me, is it ready to be merged?

piecyk:
Yes, will do 👍

PR #254, opened 2018-12-10
Fix readme regarding fragments

jmakGH:
The readme currently states users with React <=15 must use a wrapping div around the Reference and Popper components. However fragments weren't added to React until v16. This change helps users with React <=16 properly set up the components and avoid the error that branches from it.

Sources
https://reactjs.org/blog/2017/09/26/react-v16.0.html#new-render-return-types-fragments-and-strings


FezVrasta:
<= 15 means 15 and lower. I think it's right

jmakGH:
You're correct, that's my mistake. The result of weary late night eyes.

PR #257, opened 2018-12-11
perf: Reduce Popper updates when placement don't change

piecyk:
ScheduleUpdate is called twice per state change, as `placement !== this.state.placement` is not inside the `setState` callback, so the `this.state` is referring to current state and it's calling `scheduleUpdate` when the callback fires

In the same time in `componentDidUpdate` we have the 
```js
if (prevState.placement !== this.state.placement) {
  this.scheduleUpdate();
}
```
state placement changed, call the `scheduleUpdate`

Outcome of this is that on every new `state.placement` we call `two` time the `scheduleUpdate` IMHO it's redundant as scheduleUpdate is called at the same time.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,25 +1,25 @@
 {
   "dist/index.umd.js": {
-    "bundled": 49749,
-    "minified": 17751,
-    "gzipped": 5578
+    "bundled": 49689,
+    "minified": 17716,
+    "gzipped": 5566
   },
   "dist/index.umd.min.js": {
-    "bundled": 24376,
-    "minified": 10020,
-    "gzipped": 3347
+    "bundled": 24316,
+    "minified": 9985,
+    "gzipped": 3336
   },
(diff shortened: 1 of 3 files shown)


piecyk:
hmm, looks like this breaks correct position of popper at my internal tooltip implementation... 
Need to look closer... will ping with results

piecyk:
Checkout what was the problem... changing initial state placement value to what we pass in props is not good for poppers with `arrow` as we set placement to child function after we got first popper data.

After first placement, arrow is added with some margin that's increase popper element height and invalidates it's position.

Current flow, 
1. first render, popper.js is triggered, state.placement = undefined 
2. we got data,  data.placement !== state.placement, trigger schedule update 
3. we got data, data.placement === state.placement, done

 initial placement on state should stay `undefined`

@FezVrasta it's ready

FezVrasta:
thanks so much! merged

Issue #260, opened 2018-12-18
Flow error for passing this.props.getReferenceRef to warning()


tamias:
### Reproduction demo

Reproducing the problem requires setting up a project with Flow.

### Steps to reproduce the problem

1.  Set up a project using react-popper and flow-typed.
2.  Run flow.

### What is the expected behavior?

No errors from Flow.

### What went wrong?

Flow reports the following error:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ node_modules/react-popper/lib/cjs/Reference.js.flow:25:7

Cannot call warning with this.props.getReferenceRef bound to shouldBeTrue because:
 • function type [1] is incompatible with boolean [2].
 • undefined [1] is incompatible with boolean [2].

     node_modules/react-popper/lib/cjs/Reference.js.flow
 [1] 14│   getReferenceRef?: (?HTMLElement) => void,
       :
     22│
     23│   render() {
     24│     warning(
     25│       this.props.getReferenceRef,
     26│       '`Reference` should not be used outside of a `Manager` component.'
     27│     );
     28│     return unwrapArray(this.props.children)({ ref: this.refHandler });

     flow-typed/npm/warning_v3.x.x.js
 [2]  5│   declare module.exports: (shouldBeTrue: boolean, warning: string) => void;
```

### Any other comments?

I believe this can be fixed simply by casting `this.props.getReferenceRef` to a boolean with `!!`.

### Packages versions

- Popper.js: 1.14.4
- react-popper: 1.3.2

FezVrasta:
Oh you are right, thanks for the report!

If you are able to send a PR it'd be so helpful, thanks!

tamias:
Thanks for the quick response!  I've created a PR.

tamias:
I made a mistake with the original PR, so I've created a new one.  (It also would have had a conflict with d4aeb529.)

PR #261, opened 2018-12-18
fix: avoid flow error for call to warning() (#260)

tamias:
https://github.com/FezVrasta/react-popper/issues/260

Passing `this.props.getReferenceRef` to `warning()` causes an error with flow-typed, because `warning()` expects a boolean as its first argument.

Avoid the error by casting `this.props.getReferenceRef` to a boolean.

### DIFF
--- a/src/Manager.js
+++ b/src/Manager.js
@@ -1,34 +1,37 @@
 // @flow
-import * as React from "react";
-import createContext, { type Context } from "create-react-context";
+import * as React from 'react';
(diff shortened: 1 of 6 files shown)


FezVrasta:
Thanks! May you also install the `warning` flow-typed definitions so we don't get these problems in future?

tamias:
Well, that didn't do what I wanted it to do...  I'll start over.

PR #262, opened 2018-12-19
fix: avoid Flow error in call to warning() (#260)

tamias:
#260 

With flow-typed, warning() expects its first argument to be a boolean.
Added tests to ensure warning() is called correctly.

### DIFF
--- a/src/Reference.js
+++ b/src/Reference.js
@@ -24,7 +24,7 @@ class InnerReference extends React.Component<
 
   render() {
     warning(
-      this.props.setReferenceNode,
+      Boolean(this.props.setReferenceNode),
       '`Reference` should not be used outside of a `Manager` component.'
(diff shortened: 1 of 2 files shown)


tamias:
Regarding installing the flow-typed definition for `warning`...  It turns out that there isn't a flow-typed definition for `warning` v4+ yet, so we'd just get an autogenerated stub that uses `any` everywhere.

We're getting that Flow error in our project because we have a direct dependency warning v3, which does have a definition in flow-typed.

So, is it still worth installing flow-typed definitions at this point?

FezVrasta:
That's fine then! I wasn't aware v3 shipped with type definitions

FezVrasta:
That's fine then! I wasn't aware v3 shipped with type definitions

PR #266, opened 2019-01-28
fix: avoid loop on Reference ref setting

FezVrasta:


src/Manager.js
```diff
@@ -31,10 +31,15 @@ export default class Manager extends React.Component<
     };
   }
 
-  setReferenceNode = (referenceNode: ?HTMLElement) =>
+  setReferenceNode = (referenceNode: ?HTMLElement) => {
+    if (!referenceNode || this.state.context.referenceNode === referenceNode) {
```
FezVrasta: It feels hacky but it's the only easy fix I found to avoid render loops when the reference element is some "complex" component

piecyk: @FezVrasta yeah, also don't seeing any other solution for this 👍 

One note, it's ok to have stale referenceNode here, but would add on `InnerReference`, 
```tsx 
componentWillUnmount() {
  safeInvoke(this.props.innerRef, null);
}
```
that would clean refs that Caller of `react-popper` to avoid stale references (and memory leaks) on Caller of react-popper. Similar as in https://github.com/FezVrasta/react-popper/blob/master/src/Popper.js#L189

FezVrasta: Thanks!

https://github.com/FezVrasta/react-popper/commit/5885b8e29f4e68979dba36d8e0e4ff987851d8a7

PR #268, opened 2019-01-29
chore: upgrade to jest@24 thus removing babel 6 dependency

mohsinulhaq:
Jest 24 uses babel 7 now. As such, babel-core bridge is no longer required.


FezVrasta:
Thanks!

PR #279, opened 2019-05-16
Fix update/re-rendering logic

kinke:
Previously, re-rendering a `<Popper />` component manually would not update the Popper instance, as long as the `placement` prop did not change and the render function returned a similar DOM tree (so that `popperNode` remained unchanged). E.g., if the render function returned a larger child, Popper.js would not get a chance to re-position the popper element accordingly.

Fix this by always recreating/updating the Popper instance in `componentDidUpdate`, and being somewhat more clever in the `updateStateModifier()` function (just updating the state and hence re-rendering if something of relevance did actually change).

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,25 +1,25 @@
 {
   "dist/index.umd.js": {
-    "bundled": 52221,
-    "minified": 17785,
-    "gzipped": 5683
+    "bundled": 52578,
+    "minified": 18111,
+    "gzipped": 5843
   },
   "dist/index.umd.min.js": {
-    "bundled": 26233,
-    "minified": 10054,
-    "gzipped": 3437
+    "bundled": 26590,
+    "minified": 10380,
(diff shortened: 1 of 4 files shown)


FezVrasta:
Thanks for the PR. This change would only cover the case where React is responsible for the size change, but wouldn't cover CSS related size changes.

A more correct approach to this is to use a resize event listener (such as `react-resize-aware`) to trigger a position updated whenever needed.

I'm sorry for the time it took you to set this PR up, but this is a design decision made time ago.

kinke:
> This change would only cover the case where React is responsible for the size change, but wouldn't cover CSS related size changes.

Yep, not only that, it also doesn't handle the case where a child inside the Popper component is re-rendered and triggers a size change. So a ResizeObserver-based solution would be a proper fix.

Still, I think this partial solution is better than the status quo, and it also reduces some superfluous re-renderings, as the touched unittest shows.

PR #281, opened 2019-06-17
Upgrade create-react-context to 0.3.0

vkrol:
`create-react-context@0.3.0` is MIT again https://github.com/jamiebuilds/create-react-context/blob/master/package.json#L9.
The related commit is https://github.com/FezVrasta/react-popper/commit/5c423ed2ee9d51511395cbbad734156d02541cbb.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,13 +1,13 @@
 {
   "dist/index.umd.js": {
-    "bundled": 52221,
-    "minified": 17785,
-    "gzipped": 5683
+    "bundled": 51005,
+    "minified": 17508,
+    "gzipped": 5553
   },
   "dist/index.umd.min.js": {
-    "bundled": 26233,
-    "minified": 10054,
-    "gzipped": 3437
+    "bundled": 25104,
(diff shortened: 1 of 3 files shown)


FezVrasta:
Could we rather replace it with https://www.npmjs.com/package/mini-create-react-context?

FezVrasta:
Thanks!

jerroydmoore:
This PR might be worth a micro release.

The current public release of react-popper depends on create-react-context@0.2.2, which depends on an older verison of fbjs, which is using an older and unmanaged verison of core-js.

`npm install` logs the following complain:
```
core-js@1.2.7: core-js@<2.6.8 is no longer maintained. Please, upgrade to core-js@3 or at least to actual version of core-js@2.
```

vsakaria:
I was sniffing around and found duplicate entries for `create-react-context` both `create-react-context@^0.2.2` and `create-react-context@^0.3.0` in the `yarn.lock` I was wonder if this was a bug on `yarn`.

Now I am thinking its intentional.

Either its note worthy.

PR #292, opened 2019-08-04
Refactor render prop descriptions and add some detail about arrow styles

ragurney:
Restructures the render prop details and adds a little more about the arrow style. 

My editor also decided to correct some markdown style, which can be left out of this PR if they're not wanted.

Stems from https://github.com/FezVrasta/react-popper/issues/291

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 React wrapper around [Popper.js](https://popper.js.org).
 
-**important note:** popper.js is **not** a tooltip library, it's a *positioning engine* to be used to build features such as (but not restricted to) tooltips.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -132,8 +126,8 @@ Function that can be used to obtain popper reference
 placement?: PopperJS$Placement;
 ```
 
-One of the accepted placement values listed in the [Popper.js documentation](https://popper.js.org/popper-documentation.html#Popper.placements).  
-Your popper is going to be placed according to the value of this property.
```
FezVrasta: 2 trailing whitespaces in markdown mean "newline" (`<br />` tag), please may you revert this?

PR #297, opened 2019-08-16
fix: update react-spring so demo runs locally

mobot11:
The current version of react-spring throws an error when the demo is compiling. Updating the dependency fixes the build. This fixes https://github.com/FezVrasta/react-popper/issues/269


mobot11:
Build is still broken. Closing for now.

PR #298, opened 2019-08-20
chores : using latest popper dependency

vasikarla:


package.json
```diff
@@ -62,7 +62,7 @@
   "dependencies": {
     "@babel/runtime": "^7.1.2",
     "create-react-context": "^0.3.0",
-    "popper.js": "^1.14.4",
+    "popper.js": "1.15.0",
```
TrySound: 1.15.0 is considered by `^1.14.4` range. If you want to upgrade transitive dependency remove and install again react-popper.

vasikarla: @TrySound : how do i handle this if im using popper dependency in my codebase?

TrySound:
```
yarn remove react-popper && yarn add react-popper
```
or
```
npm uninstall react-popper && npm install react-popper
```

FezVrasta:
Closing as per the comments above

package.json
vasikarla: let me try that!

PR #306, opened 2019-09-11
Fix: Allow custom arrow and applyStyle modifiers

dmitriirudakov:
Please refer to the next bug to find some more information about these changes.
https://github.com/FezVrasta/react-popper/issues/305


FezVrasta:
I think this has been fixed with #333

PR #308, opened 2019-09-16
Allow ref objects to be passed

### LINKED ISSUE #307: Allow ref objects
Right now react-popper only supports ref callbacks, it should also support ref objects so folks can use hooks and `React.useRef` with this library.

maclockard:
Fixes #307 

I tested this locally with my app using `yarn link` and everything worked!


typings/react-popper.d.ts
```diff
@@ -6,36 +6,37 @@ interface ManagerProps {
...
 export class Manager extends React.Component<ManagerProps, {}> { }
 
-type RefHandler = (ref: HTMLElement | SVGElement | null) => void;
-
 interface ReferenceChildrenProps {
-  ref: RefHandler;
+  // React refs are supposed to be contravariant (allows a more general type to be passed rather than a more specific one)
+  // However, Typescript currently can't infer that fact for refs
+  // See https://github.com/microsoft/TypeScript/issues/30748 for more information
+  ref: React.Ref<any>;
```
maclockard: this is very unfortunate, but this issue with contravariance vs covariance with refs is a very long standing issue with TS that doesn't look like it will be resolved soon.

src/RefTypes.js
```diff
@@ -0,0 +1,4 @@
+type RefHandler = (?HTMLElement) => void;
+type RefObject = { current?: HTMLElement};
+
+export type Ref = RefHandler | RefObject;
```
FezVrasta: Why can't we use `React$Ref<HTMLElement>` here?

piecyk:
@maclockard would be awesome if you add test for this 👍

src/RefTypes.js
maclockard: [`React$Ref<HTMLElement>`](https://flow.org/en/docs/react/types/#toc-react-ref) includes string refs which are legacy and react-popper does not support.

maclockard:
Added tests!

Ended up needing to bump some of the test versions to properly support ref objects, so theres a bit more churn on the snapshots than I would like, but that's needed for https://github.com/FezVrasta/react-popper/pull/303 anyway.

Still something a little bit funky with the flow typings since it doesn't recognize `React.createRef` even though it exists and works, but not too concerned since its only used in tests. I'm a bit more of a typescript person so if you have thoughts, let me know.

maclockard:
@FezVrasta rebased this PR on top of master, could you take a look?

maclockard:
@FezVrasta do you think you will have time to look at this soon?

shepmaster:
I have code like this:

```ts
      <Reference>
        {({ ref }) => <Button ref={(r) => { ref(r); buttonRef.current = r; }} toggle={toggle} />}
      </Reference>
```

Which now reports an error:

```
[tsl] ERROR in PopButton.tsx(39,45)
      TS2349: This expression is not callable.
  Not all constituents of type 'Ref<any>' are callable.
    Type 'RefObject<any>' has no call signatures.
```

Is there a suggested change I should make?

maclockard:
@shepmaster how are you declaring ref? Are you using `React.useRef` or `React.createRef` or are you are you manually creating a callback ref? Also what versions of react and react popper are you using? Is this in a function component or a class component?

shepmaster:
@maclockard sorry for the initial lack of information!

- React: 16.12.0
- React Popper: 1.3.4 (attempting to update to 1.3.5)
- This is in [a functional component](https://github.com/integer32llc/rust-playground/blob/master/ui/frontend/PopButton.tsx#L12-L64)
- `buttonRef` is from `useRef<HTMLElement>()`;

[The whole project](https://github.com/integer32llc/rust-playground/tree/b39aa11261defee57bee81c4442f4491637dc384/ui/frontend).

The code currently [*appears* to work](https://play.rust-lang.org/) (e.g the "Debug" menu button).

Thanks!

maclockard:
@shepmaster the code you have is correct in the sense that it will run, but its just failing a type check. The `ref` you are being passed is a callback ref in actuality, but the type is [`React.Ref`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L89) which is a union type of a callback ref and a ref object. So Typescript doesn't know which variant `ref` is and throws a type error since you can't call a ref object.

The best way to fix this is for you to add a [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types) to provide TS with enough information to know that `ref` is a callback ref and not a ref object. Here's an example of doing this:

```
function isRefObject<T>(ref: React.Ref<T>): ref is RefObject<T> {
  return ref !== null && typeof ref !== 'function';
}
```

The other (worse) solution would be to change the [type of `ref`](https://github.com/popperjs/react-popper/blob/master/typings/react-popper.d.ts#L23) in react-popper to just be a callback ref instead of `React.Ref`. I personally do not like this solution since it would couple react-popper's API to be tied to using callback refs, meaning it would be a breaking change for react-popper to start using `React.createRef` or `React.useRef` in the future. By keeping the more general `React.Ref` type, react-popper becomes more future proof. By adding a type guard for this case, you code becomes more defensive in the case react-popper changes its ref type.

maclockard:
Another solution is that react-popper provides an [`innferRef` prop](https://github.com/popperjs/react-popper/blob/master/typings/react-popper.d.ts#L18) on `Reference`. You could pass `buttonRef` as that prop and react-popper will correctly combine `buttonRef` and the ref it uses under the hood. Here's the code that does that: https://github.com/popperjs/react-popper/blob/master/src/Reference.js#L21-L24.

So your code would look like this:
```
      <Reference innerRef={buttonRef}>
        {({ ref }) => <Button ref={ref} toggle={toggle} />}
      </Reference>
```

shepmaster:
> Another (probably better) solution is that react-popper provides an [`innferRef` prop](https://github.com/popperjs/react-popper/blob/master/typings/react-popper.d.ts#L18) on `Reference`

Ah, that works wonderfully, thank you! For future readers, there's [also an `innerRef` on `Popper`](https://github.com/popperjs/react-popper/blob/eafe0cc7f665aea9118175f520be96441081ee8e/typings/react-popper.d.ts#L39).

Now I'll have to wonder if those were always there and I missed them on my original implementation or what... 😉

PR #309, opened 2019-09-16
Set side effects to 'false'

maclockard:
This sets the `package.json` field `sideEffects` to `false`. This allows for better tree shaking by downstream consumers, see https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free for more information.

I skimmed over all included files and I did not see anything that would count following webpack's definition of tree shaking.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   "main": "lib/cjs/index.js",
   "module": "lib/esm/index.js",
   "typings": "typings/react-popper.d.ts",
+  "sideEffects": false,
   "files": [
     "/dist",
     "/lib",


shlomitc:
@maclockard  - Did you test how it affected your bundle size? was it reduced significantly?

maclockard:
@shlomitc This was a bit ago so I don't quite remember exactly what the improvement was. I'm betting it was on the smaller side since react-popper doesn't add that much to size itself, it's more from its dependencies ([link](https://bundlephobia.com/result?p=react-popper@1.3.7)) and popper.js is kinda all or nothing from a shaking perspective. 

I generally try to do this for all dependencies I use, big or small, since tree shaking has a great effect cumulatively.

 If you want to test the diff yourself, I usually test using either using [bundlesize](https://github.com/siddharthkp/bundlesize) for a quick overview and comparing to past sizes or [webpack-bundle-analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer) for a more in depth exploration or what exactly is included/excluded.

PR #313, opened 2019-10-02
fix: referenceNode memory leak in Manager (#312)

### LINKED ISSUE #312: Memory leak when Reference children are removed from a page
### Reproduction demo

https://codesandbox.io/s/react-popper-v1x-issue-template-3hcp9

1.  Replace the demo code in the repo's `demo/index.js` file with the code in `index.js` of the [codesandbox's index.js file](https://codesandbox.io/s/react-popper-v1x-issue-template-3hcp9) 
2.  Start the demo app and load it in Chrome
3.  Open Chrome devtools and take a heap snapshot
4.  Click on "Toggle Popout" in the demo page
5.  Take another heap snapshot
6.  In the second heap snapshot look for "Detached HTMLDivElement" with the shortest distance and find in the retainer panel that `referenceNode` is retaining the div element 

`referenceNode` in `Manager` is set to `null` when the Reference element is removed from the page.

### What went wrong?

When the Reference element is removed from the page, the `ref` value passed to the Manager context is retained. This creates a memory leak in situations where the reference element is regularly removed from a page.

### Any other comments?

- This is most easily seen in a production build of React since the dev build adds a lot of extra utilities that prevents a clear exposure of the leak.

### Packages versions

- Popper.js: whatever version react-popper 1.3.4 depends on
- react-popper: 1.3.4

linusthe3rd:
Fixes #312 

Instead of storing referenceNode in state, store it in a class property. This is done to enable the component to cleanup referenceNode’s value during unmounting.

I also added unit tests to account for this change, but that resulted in me having to update some dependencies and snapshots to get unit tests to pass.


package.json
```diff
@@ -101,8 +101,8 @@
     "parcel-bundler": "^1.10.1",
     "prettier": "^1.11.1",
     "pretty-quick": "^1.4.1",
-    "react": "16.2.0",
-    "react-dom": "^16.3.0",
+    "react": "16.9.0",
+    "react-dom": "^16.9.0",
```
linusthe3rd: The reason for this update is to get unit tests to pass. I basically copied what was done in this PR: https://github.com/FezVrasta/react-popper/pull/308/commits/a86c9c6366c44acd925443aa65f1f8364900739a

src/Manager.js
```diff
@@ -2,48 +2,50 @@
...
 
   setReferenceNode = (referenceNode: ?HTMLElement) => {
-    if (!referenceNode || this.state.context.referenceNode === referenceNode) {
+    if (!referenceNode || this.contextObj.referenceNode === referenceNode) {
       return;
     }
 
-    this.setState(({ context }) => ({
-      context: { ...context, referenceNode },
-    }));
```
linusthe3rd: I had to remove the usage of `state` because React doesn't want `state` to be updated in `componentWillUnmount`. To get around this, I utilized class attributes to prevent new objects being created for the `value` property and `forceUpdate` to trigger a re-render when the `referenceNode` is updated.

From everything I can tell based on the git history, I _think_ this continues to support the use cases for this class, but happy to evaluate other approaches if I am missing something.

piecyk: Yes this is correct approach, storing html element in state was not nice. Good work 👍 

We can also split ManagerContext into two, one responsible for setting the element and another for passing the element... Then when Reference component will call ref it will no be re-rendered.

Something like this https://codesandbox.io/s/popper-context-lig79

linusthe3rd: That's a good call - it negates the constraint of React always re-rendering consumers of the context for new objects.

linusthe3rd: Updated

src/Manager.js
```diff
@@ -2,50 +2,38 @@
...
-      },
-    };
-  }
+export default class Manager extends React.Component<ManagerProps> {
+  referenceNode: ?HTMLElement;
 
-  setReferenceNode = (referenceNode: ?HTMLElement) => {
-    if (!referenceNode || this.state.context.referenceNode === referenceNode) {
+  setReferenceNode = (newReferenceNode: ?HTMLElement) => {
+    if (!newReferenceNode || this.referenceNode === newReferenceNode) {
```
piecyk: Imho here we can go with 

```ts
if (this.referenceNode !== newReferenceNode) {
  this.referenceNode = newReferenceNode;
  this.forceUpdate();
} else {
  this.referenceNode = newReferenceNode;
}
```

Then we don't need to clear the  `referenceNode ` in `componentWillUnmount` and  Manager component will always have current value of reference node

src/Manager.js
```diff
@@ -2,50 +2,38 @@
 import * as React from 'react';
 import createContext, { type Context } from 'create-react-context';
 
-export const ManagerContext: Context<{
-  setReferenceNode?: (?HTMLElement) => void,
-  referenceNode?: ?HTMLElement,
-}> = createContext({ setReferenceNode: undefined, referenceNode: undefined });
+export const ManagerReferenceNodeContext: Context<?HTMLElement> = createContext();
+export const ManagerReferenceNodeSetterContext: Context<(?HTMLElement) => void> = createContext();
```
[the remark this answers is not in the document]
linusthe3rd: fixed.

src/Manager.js
linusthe3rd: I updated to use this code, but would this recreate the issue mentioned in this comment in the original code at all? 

https://github.com/FezVrasta/react-popper/pull/266/files#r251516842

I don't think so because the problem implied there was `context` forcing a re-render due to a complex object being passed down to children, but want to double check.

piecyk: Exactly 👍

piecyk:
@FezVrasta please have a look on this when you got some time, thanks 👍

linusthe3rd:
@FezVrasta I've resolved the merge conflicts on the PR. LMK if there's anything else I can do to help get this merged in!

FezVrasta:
Sorry guys, I'll review/merge it as soon I get a chance. Thank you for your help with this PR!

andreyvital:
@FezVrasta any updates? ☺️ 😅

linusthe3rd:
For those following this PR, I was able to implement a workaround to this leak by utilizing the [`referenceElement` API](https://github.com/FezVrasta/react-popper/blob/master/src/Popper.test.js#L21-L22) on the `Popper` component. 

The one caveat with this is if you want the Popper to render as open at the same time as the `referenceElement`, you may encounter a bug where the popout is not positioned correctly. This occurs when `referenceElement` is a `ref` and `current` has not been set yet. To get around this, I created a callback ref to track when my `referenceElement` ref gets updated and force a re-render.

For example:

```js
const targetRed = useRef();
const [shouldRenderPopout, setShouldRenderPopout] = useState(false);
const childrenRef = el => {
  targetRef.current = el;
  if (targetRef.current) {
    setShouldRenderPopout(true);
  }
};

return (
  <Fragement>
    <div ref={childrenRef}>Foo</div>
    <Popper referenceElement={targetRef.current} {...props}>
      {/* ...stuff... */}
    </Popper>
  </Fragment>
);
```

This is [effectively what the `Manager` component is doing](https://github.com/FezVrasta/react-popper/blob/master/src/Manager.js#L34-L49)- the `context` API it uses triggers a re-render when its value is updated due to the `Object.is` check that occurs on `context` changes.

andreyvital:
Great @linusthe3rd. Thanks for sharing. But I guess I could use your PR's fix with https://github.com/ds300/patch-package

linusthe3rd:
The latest merge conflicts have been resolved

Issue #318, opened 2019-11-13
Make react-popper up-to-date with flow-bin


debel27:
I'm unable to upgrade flow-bin to "^0.111.0" in my project because it triggers errors in react-popper.

Would it be possible to upgrade the flow-bin dependency in react-popper to match the latest flow version ? I can attempt a PR for this, if you're willing to accept it

### Reproduction demo

See next section

### Steps to reproduce the problem

```
mkdir test && cd test
npm init -y
npm i react react-popper flow-bin
printf "[include]\nsrc/**/*.js" > .flowconfig
mkdir src
printf "//@flow\nimport { Popper } from 'react-popper';" > src/index.js
npx flow

```

### What is the expected behavior?

No flow errors 

### What went wrong?

I got the following flow errors

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ node_modules/react-popper/lib/cjs/Popper.js.flow:120:14

Cannot determine a type for object literal [1]. CSSStyleDeclaration [2] cannot be spread because the indexer number [3]
may overwrite properties with explicit keys in a way that Flow cannot track. Can you spread CSSStyleDeclaration [2]
first or remove the indexer?

     node_modules/react-popper/lib/cjs/Popper.js.flow
     117│       ? initialStyle
 [1] 118│       : {
     119│           position: this.state.data.offsets.popper.position,
     120│           ...this.state.data.styles,
     121│         };
     122│
     123│   getPopperPlacement = () =>

     /tmp/flow/flowlib_11180f0/cssom.js
 [3] 408│   [index: number]: string;

     node_modules/popper.js/dist/umd/popper.js.flow
 [2]  41│   styles: CSSStyleDeclaration,


Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ node_modules/react-popper/lib/cjs/Popper.js.flow:233:15

Cannot determine a type for props [1]. rest of object pattern [2] is inexact, so it may contain referenceElement with a
[...]

PR #319, opened 2019-11-13
fix: upgrade to flow-bin ^0.112.0

debel27:
This is my first attempt at fixing #318 

Currently, I only fixed what broke after the upgrade flow-bin. But I have some comments :

- I've used exact types for the props of Popper and Reference. I think it's not a problem, because there doesn't appear to be any spreading whatsoever. Is it acceptable that I migrate the props types of other components, so they are exact as well ? For the sake of API consistency
- In `Popper.js`, the `position` style cannot be overwritten anymore by what's in `state.data.styles`. Is that OK ?

### DIFF
--- a/src/Popper.js
+++ b/src/Popper.js
@@ -30,7 +30,7 @@ export type PopperChildrenProps = {|
 |};
 export type PopperChildren = PopperChildrenProps => React.Node;
 
-export type PopperProps = {
+export type PopperProps = {|
   children: PopperChildren,
   eventsEnabled?: boolean,
   innerRef?: Ref,
@@ -38,12 +38,12 @@ export type PopperProps = {
   placement?: Placement,
   positionFixed?: boolean,
   referenceElement?: ReferenceElement,
-};
+|};
 
-type PopperState = {
+type PopperState = {|
   data: ?Data,
   placement: ?Placement,
-};
+|};
 
 const initialStyle = {
   position: 'absolute',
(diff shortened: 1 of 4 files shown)


src/Popper.js
```diff
@@ -116,8 +116,8 @@ export class InnerPopper extends React.Component<PopperProps, PopperState> {
     !this.popperNode || !this.state.data
       ? initialStyle
       : {
-          position: this.state.data.offsets.popper.position,
           ...this.state.data.styles,
+          position: this.state.data.offsets.popper.position,
```
FezVrasta: Isn't `data.offsets.popper.position` an exact type? What error do we get if we keep it where it was before?

debel27: The actual issue is with the type of `this.state.data.styles`. Its Flow type is `CSSStyleDeclaration` (from the CSSOM Flow types), which happen to have `number` as keys. 

For some reason, this makes Flow unsure about whether `position` would be overwritten by the spread or not. Moving `position` after the spread removes ambiguity.

Here is the original Flow error :

```
Error --------------------------------------------------------------------------------------------- src/Popper.js:120:12

Cannot determine a type for object literal [1]. `CSSStyleDeclaration` [2] cannot be spread because the indexer
number [3] may overwrite properties with explicit keys in a way that Flow cannot track. Can you spread
`CSSStyleDeclaration` [2] first or remove the indexer?

   src/Popper.js:120:12
   120|         ...this.state.data.styles,
                   ^^^^^^^^^^^^^^^^^^^^^^

References:
   src/Popper.js:118:9
                v
   118|       : {
   119|           position: this.state.data.offsets.popper.position,
   120|         ...this.state.data.styles,
   121|         };
                ^ [1]
   node_modules/popper.js/index.js.flow:39:13
    39|     styles: CSSStyleDeclaration,
                    ^^^^^^^^^^^^^^^^^^^ [2]
   ../../AppData/Local/Temp/flow/flowlib_218c75f5/cssom.js:408:11
   408|   [index: number]: string;
                  ^^^^^^ [3]
```

debel27: @FezVrasta, is the change OK for you ?

PR #320, opened 2019-11-22
Fix infinite loop

mfedderly:
Popper.js' componentDidUpdate fails the props change test, which calls updatePopperInstance, which eventually causes componentDidUpdate to get called again. By using a deepEquals comparison on props.modifiers we correctly determine that the props did not update, and break the loop.

I don't have a nice simple repro for this issue, but it seems to only trigger in a complicated app on Chrome (at least 70.x->78.x) on Windows when the page's zoom level is less than 100%.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,25 +1,25 @@
 {
   "dist/index.umd.js": {
-    "bundled": 67296,
-    "minified": 23836,
-    "gzipped": 7176
+    "bundled": 88323,
+    "minified": 32847,
+    "gzipped": 9575
   },
   "dist/index.umd.min.js": {
-    "bundled": 31932,
-    "minified": 12958,
-    "gzipped": 4186
+    "bundled": 52955,
+    "minified": 21926,
+    "gzipped": 6619
   },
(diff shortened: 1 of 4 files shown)


DanHarman:
I have this happening in our app and for us it only happens at precisely 75% zoom in chrome. We are using a blueprint datepicker but I can't repro on their test page.

mfedderly:
@FezVrasta any chance you could look at this PR?

jalessio:
This patch resolves reproducible browser freeze (CPU in busy loop) in our app which uses Blueprint. My reproducible test case involves running Chrome on Windows at 75% zoom (as mentioned by @DanHarman above). 

I will note that we are also getting reports of similar issues at 100% zoom. I'm hopeful this fix will help there as well, but I haven't been able to reliably reproduce that issue yet.

styu:
Any chance we can get a patch release with this fix? We'd love to update to fix the freeze issues we've been seeing in our apps!

FezVrasta:
1.3.7, there you go.

Thanks again for the help everybody!

schlosna:
@FezVrasta thank you!

PR #326, opened 2020-01-07
Export popper.js typings

Guillaumez:
Export `popper.js` types from `react-popper` to avoid the `import/no-extraneous-dependencies` error from ESLINT when we need `popper.js` typings.


FezVrasta:
If you need to access Popper.js types you should add the library as a dependency, it doesn't make sense to proxy it this way, it would lead to issues down the road.

PR #327, opened 2020-01-13
docs: add missing header in README

lluia:
When going through the `README.md` I noticed there was a missing header for the `outOfBoundaries` documentation, which made it look as it was part of `placement` documentation 🧐 💅🏻


FezVrasta:
Thanks!

PR #333, opened 2020-02-02
feat: migration to popper v2

sgrishchenko:
**Migration Guide**
1. @popperjs/core is peer dependency now. Instalation: 
   ```shell
   # With npm
   npm i react @popperjs/core react-popper

   # With Yarn
   yarn add react @popperjs/core react-popper
   ```
   If you want use `placements` or `modifierPhases` enums, you can import they directly from [@popperjs/core](https://popper.js.org/docs/v2/tree-shaking/):
   ```js
   import { placements, modifierPhases } from '@popperjs/core/lib';
   ```
1. `modifiers` prop should be present as array with changed format (see [official docs](https://popper.js.org/docs/v2/migration-guide/#10-update-modifiers)).

1. `eventsEnabled` prop is removed, you can use [Event Listeners](https://popper.js.org/docs/v2/modifiers/event-listeners/) modifier to control it  (see [official docs](https://popper.js.org/docs/v2/migration-guide/#8-update-method-names)).

1. `positionFixed` prop is removed, you can use `strategy` prop instead (see [official docs](https://popper.js.org/docs/v2/migration-guide/#7-change-positionfixed)).

1. `outOfBoundaries` is not provided in `<Popper>` component render child anymore, two new props are provided now: `isReferenceHidden` and `hasPopperEscaped`. They are provided from [Hide](https://popper.js.org/docs/v2/modifiers/hide/) modifier.

1. `scheduleUpdate` method is renamed to `update` (like in [@popperjs/core](https://popper.js.org/docs/v2/migration-guide/#8-update-method-names)). Now it returns `Promise<PopperJS.State>` or `Promise<null>` if you try call it when `Popper Instance` is not defined.

**Change List**
- [x] support new API
- [x] provide new API for `<Popper>` component
- [x] update Demo
- [x] update Typesctipt typings
- [x] update unit tests
- [x] fix flow errors
- [x] fix typescript errors
- [x] write migration guide


package.json
```diff
@@ -58,13 +58,13 @@
     "popover"
   ],
   "peerDependencies": {
+    "@popperjs/core": "^2.0.0",
```
FezVrasta: This shouldn't be here, having it in `dependencies` is enough.

src/Popper.js
```diff
@@ -1,21 +1,25 @@
...
+  type PositioningStrategy,
   type Instance,
-  type Data,
-  type Modifiers,
-  type ReferenceObject,
-} from 'popper.js';
+  type VirtualElement,
+  type Modifier,
+  type ModifierArguments,
+} from '@popperjs/core/lib';
```
FezVrasta: I think importing from `@popperjs/core` should be enough

atomiks:
Isn't v2 just going to be a `usePopper` hook?

FezVrasta:
I was thinking the same, but probably it won't hurt to first upgrade the existing API to use Popper 2, and later add the `usePopper` API to the library as a feature update?

src/Popper.js
sgrishchenko: unfortunately it doesn't work, popper exports from main file `@popperjs/core` only three things: 
```js
export { createPopper, popperGenerator, defaultModifiers };
```

sgrishchenko: See [this one](https://github.com/popperjs/popper-core/blob/49d2ee9bb6d58f43ca1d5183095a1af253fe9aa7/src/popper.js#L28) + [official doc](https://popper.js.org/docs/v2/typings/)

package.json
sgrishchenko: This is common practice (see [react-redux](https://github.com/reduxjs/react-redux/blob/master/package.json#L40) or [mobx-react](https://github.com/mobxjs/mobx-react/blob/master/package.json#L44)). And now we don't need re-export things like `placements` of `strategies`. User can import all that he or she need directly from original library.

akozhemiakin:
First of al thanks for the great lib. When Popper 2 is release we implemented thin react integration locally in our project and I extracted it into one gist (in case there is any chance it could be usefull):

https://gist.github.com/akozhemiakin/e91c23b8ebc2ceb9c59edab947b2de92

ksocha:
> I was thinking the same, but probably it won't hurt to first upgrade the existing API to use Popper 2, and later add the `usePopper` API to the library as a feature update?

@FezVrasta 

Adding a hook to the lib would most likely force to change peerDependency
`"react": "0.14.x || ^15.0.0 || ^16.0.0"` to `"react": "^16.8.0"` which IMO should be considered as a breaking change.

Additionally, adding a hook later means that there will be two implementations of the popper in the lib: hook-based and class-based. It would be more reasonable to create a hook-based implementation first and reuse it to create functional component with render prop to replace current class-based implementation.

Thank you for your hard work, great lib!

sgrishchenko:
Hi, folk. I created this PR in hope that we can iteratively upgrade this library to Popper v2. Honestly, I don't understand why we can't merge this one and move on. 

My changes don't relates to hook API. Hook API can be added in next iterations. These changes don't mean that there will be two implementations: hook-based and class-based. We can always move core implementation to hooks and add wrappers for supporting previous API. And hook API is not silver bullet. How will developers who prefer classes use this library, if there is only hook API? Should they write own wrappers?

@FezVrasta If you disagree with my answers for your comments, just tell me, and I will fix all.

I think, new API providing is very complex problem and we need some small milestones to achieve main goal. Let's try to take small steps towards right direction. This will help library user to get use to to new API. We can release first implementation, get feedback and continue improve this solution.

FezVrasta:
I’m sorry if it’s taking so long, I’ve been through some major changes in my life and now both my wife’s and mine families are locked down in Italy, we live elsewhere so we are very worried and this doesn’t help me to focus on my side projects.

I’ll try to get back to this PR in the next days.

sgrishchenko:
Oh, sorry if I was too persistent. I sincerely wish a speedy resolution to all your problems. I hope that I can help with the development as much as I can.

PR #336, opened 2020-02-18
fix: change initial position style based on positionFixed prop

zkuzmic:
I have a scenario where I am autofocusing an element appearing inside of a popper, and that is causing the page to scroll to the top when initially opened. A similar issue can be found here:

https://github.com/mui-org/material-ui/issues/16740

Their solution was to add `position: fixed` to the initial styles on the popper, but this library doesn't offer a way to do that. This PR makes it so that initial position style will either be `absolute` or `fixed` based on the `positionFixed` prop provided to the Popper component.

### DIFF
--- a/src/Popper.js
+++ b/src/Popper.js
@@ -47,7 +47,6 @@ type PopperState = {
 };
 
 const initialStyle = {
-  position: 'absolute',
   top: 0,
   left: 0,
   opacity: 0,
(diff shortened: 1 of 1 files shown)


zkuzmic:
Quoting a comment from @atomiks on https://github.com/popperjs/popper-core/pull/1009

> The user should always call `.focus()` after the popper was positioned, never before. They can also set `visibility` themselves too.

That makes total sense to me, although there doesn't appear to be a way to do that with the current `react-popper` api. Or am I missing something?

zkuzmic:
@FezVrasta any suggestions for this PR or anything I can do to help get this merged?

PR #343, opened 2020-04-06
feat: usePopper Hook

FezVrasta:
```js
// @flow
import * as React from 'react';
import { usePopper } from 'react-popper';

const App = () => {
  const [referenceElement, setReferenceElement] = React.useState<?Element>(null);
  const [popperElement, setPopperElement] = React.useState<?HTMLElement>(null);
  const [arrowElement, setArrowElement] = React.useState<?HTMLElement>(null);

  const { styles, attributes } = usePopper(referenceElement, popperElement, {
    modifiers: [
      { name: 'arrow', options: { element: arrowElement } },
    ],
  });

  return (
    <>
      <button ref={setReferenceElement}>Reference Element</button>
      <div ref={setPopperElement} style={styles.popper} {...attributes.popper}>
        Popper Element
        <div ref={setArrowElement} style={styles.arrow} />
      </div>
    </>
  );
}
```

CodeSandbox demo:
https://codesandbox.io/s/cold-night-u1oji


atomiks:
Can you show an example/demo of this hook in practice?

FezVrasta:
@atomiks I added an example on the first post

bengry:
1. Why is this example not using `React.useRef` and instead uses `React.useState`?
2. Maybe do something similar to downshift in regards to the props-to-be-spread API, something like this:

```js
// @flow
import { usePopper } from "react-popper";

const App = () => {
  const [referenceElement, setReferenceElement] = React.useState<?Element>(
    null
  );
  const [popperElement, setPopperElement] = React.useState<?HTMLElement>(null);
  const [arrowElement, setArrowElement] = React.useState<?HTMLElement>(null);

  const { styles, attributes } = usePopper(referenceElement, popperElement, {
    modifiers: [{ name: "arrow", options: { element: arrowElement } }],
  });

  return (
    <>
      <button ref={setReferenceElement}>Reference Element</button>
      <div
        ref={setPopperElement}
        // If you want to specify additional props (e.g. style) - pass them in when calling `getPopperProps()`. Otherwise this is optional.
        // Also, no need to for the `styles`
        {...attributes.getPopperProps({ style: { background: "red" } })}
      >
        Popper Element
        <div ref={setArrowElement} style={styles.arrow} />
      </div>
    </>
  );
};

```

FezVrasta:
> 1. Why is this example not using `React.useRef` and instead uses `React.useState`?

https://github.com/popperjs/react-popper/issues/241#issuecomment-591411605

> 2. Maybe do something similar to downshift in regards to the props-to-be-spread API, something like this [...]

Merging props wouldn't add a lot in our case, the values are all strings, so doing `{ ...yourStyles, ...styles.popper }` would do exactly the same.

bengry:
> > 1. Maybe do something similar to downshift in regards to the props-to-be-spread API, something like this [...]
> 
> Merging props wouldn't add a lot in our case, the values are all strings, so doing `{ ...yourStyles, ...styles.popper }` would do exactly the same.

Right, but it allows adding additional inputs in the future, should any be added [without breaking the API], and also - it means you only need to hook one thing up to use popper (call `getPopperProps()`), not two - (spread `attributes.popper` + passing in `styles.popper`).

FezVrasta:
attributes and styles are separated because I want to let the user decide what to apply

bengry:
@FezVrasta I've updated my ealier comment with how to solve this, while keeping the [minimal usable] API lean (since passing it the popper styles is required for the `div` to be positioned correctly).

> ```js
> <div {...attributes.popper} />
> 
> // vs.
> 
> <div {...attributes.popper} styles={{...attributes.popper.styles, { background: 'red' }} />
> ```

FezVrasta:
> ```
> <div {...attributes.popper} styles={{...attributes.popper.styles, { background: 'red' }} />
> ```

Isn't the above example basically the same number of characters as the following?

```js
<div {...attributes} styles={{ ...styles.popper, background: 'red' }} />
```

bengry:
Is it, but that's the uncommon case I think. The more common use would be: `<div {...attributes.popper} />`. I don't find myself passing an inline `style` to the popper node very often (ever?). Even if I need to style it, I'd usually use `className`.

FezVrasta:
The reasoning behind this data structure is that it mimics the one exposed by Popper, so it doesn't add an additional abstraction layer on top of it. As I explained in different occasions, react-popper is meant to be an as thin as possible layer on top of Popper, for more complex use-cases I'd leverage on wrappers on top of it (like react-tippy)

src/usePopper.js
```diff
@@ -0,0 +1,100 @@
...
+          styles: Object.fromEntries(
+            elements.map((element) => [element, state.styles[element] || {}])
+          ),
+          attributes: Object.fromEntries(
+            elements.map((element) => [element, state.attributes[element]])
+          ),
+        });
+      },
+      requires: ['computeStyles'],
+      options: { setState },
```
atomiks, line 59: What's the purpose of this?

atomiks:
Looks great imo

src/usePopper.js
FezVrasta: Nothing, I have to remove it, lol

src/usePopper.js
```diff
@@ -0,0 +1,99 @@
...
+            elements.map((element) => [element, state.attributes[element]])
+          ),
+        });
+      },
+      requires: ['computeStyles'],
+    }),
+    [setState]
+  );
+
+  useLayoutEffect(() => {
```
atomiks, line 63: `useLayoutEffect` needs to check browser env check to avoid the SSR warning

FezVrasta: 🤦‍♂️

src/usePopper.js
```diff
@@ -0,0 +1,99 @@
...
+  const updateStateModifier = useMemo(
+    () => ({
+      name: 'updateState',
+      enabled: true,
+      phase: 'write',
+      fn: ({ state }) => {
+        const elements = Object.keys(state.elements);
+
+        setState({
+          styles: Object.fromEntries(
```
atomiks, line 50: I don't think we should use `fromEntries` due to old browsers, keep polyfills as minimal as `@popperjs/core`

src/usePopper.js
```diff
@@ -0,0 +1,99 @@
...
+) => {
+  const popperInstanceRef = useRef();
+  const [state, setState] = useState<State>({
+    styles: {
+      popper: { position: 'absolute', left: '0', top: '0' },
+    },
+    attributes: {},
+  });
+  const createPopper = useMemo(
+    () => options.createPopper || defaultCreatePopper,
```
atomiks, line 32: Does `defaultCreatePopper` get tree-shaken?

src/usePopper.js
```diff
@@ -0,0 +1,99 @@
...
+
+export const usePopper = (
+  referenceElement: Element,
+  popperElement: HTMLElement,
+  options: Options = {}
+) => {
+  const popperInstanceRef = useRef();
+  const [state, setState] = useState<State>({
+    styles: {
+      popper: { position: 'absolute', left: '0', top: '0' },
```
atomiks, line 27: `position: options.strategy` (https://github.com/popperjs/popper-core/pull/1055)

src/usePopper.js
FezVrasta: I doubt it's going to be an issue, those React libraries are usually consumed through webpack+babel

src/usePopper.js
FezVrasta: No, but at least users can pass their own `createPopper` function if needed

FezVrasta:
I pushed some work to convert the render-props implementation to use hooks, it's half broken, I'd appreciate some help to get it working.

You can see the issues by running `yarn demo:dev`

FezVrasta:
Ok so we are almost there, there's just one problem, the poppers are not positioned (they default to `top: 0, left: 0`).

@atomiks do you have time to take a look? I'm not quite sure what's going on 😓

atomiks:
It's a stale dependency problem w/ `useMemo`

This fixes it:

```js
  const childrenProps = React.useMemo(
    () => ({
      ref: setPopperElement,
      style: styles.popper,
      placement: state ? state.placement : placement,
      hasPopperEscaped: state
        // need optional ?. here
        ? state.modifiersData.hide?.hasPopperEscaped
        : null,
      isReferenceHidden: state
        // and here
        ? state.modifiersData.hide?.isReferenceHidden
        : null,
      arrowProps: {
        style: styles.arrow,
        ref: setArrowElement,
      },
      forceUpdate: forceUpdate || NOOP,
      update: update || NOOP_PROMISE,
    }),
    [setPopperElement, state]
  );
```

FezVrasta:
Thanks! I added the eslint rule to make sure it doesn't happen again.

FezVrasta:
So, I suppose the PR is ready for a final review.

src/usePopper.js
atomiks: pinging this again

FezVrasta: I'm wondering if we could follow the React team advice where they suggest not to run the code at all on server-side?

FezVrasta:
Tests fixed, I replaced Enzyme with react-testing-library since it seems to work better with modern React versions. I also added tests for the new hook.

Imballinst:
hi @FezVrasta, thank you for this amazing work! Question: why does in the CodeSandbox demo, the `div` for arrow is located below the popper instead of above the popper? Is this a bug from https://github.com/popperjs/popper-core?

I also face the same thing in my local, but I don't know whether it's because of this lib or because of upstream repository. I also don't quite know whether I'm configuring something wrong, though I copy-pasted exactly the same code like in the documentation for `usePopper`. Thanks again!

atomiks:
@Imballinst Popper only positions the arrow along a single axis in order to center it to the reference. For vertical placements it's the `x` axis. It's your responsibility to position it on the other axis in CSS.

FezVrasta:
I’m sorry I’m from mobile so I can’t look into it, but  make sure you are applying the correct CSS to position the arrow. Popper provides only the offset to center it. You can look at the popper.js.org tutorial

Imballinst:
@atomiks @FezVrasta Ah, okay! Got it. That should be enough as a clue for me. Thanks and sorry for the noise. Have a nice weekend!

pleerock:
shouldn't `useRef` be used instead of `useState` for element references?

FezVrasta:
`useRef` allows to react when the reference element changes identity.

yuriybelike:
Is the demo's 

> // FIXME: we probably need some logic in the hook to detect when the provided elements are `null`

still applicable? I'm having some issues setting `referenceElement` to null in my code

FezVrasta:
> Is the demo's 
> 
> 
> 
> > // FIXME: we probably need some logic in the hook to detect when the provided elements are `null`
> 
> 
> 
> still applicable? I'm having some issues setting `referenceElement` to null in my code.

It should work with null elements

sarathms:
The codesanbox example linked above doesn't seem to work anymore. This is what I see:


How to get it to work?

FezVrasta:
It's working, it's just a silly example where the popper sits below the button 😅

PR #344, opened 2020-04-08
feat: replace `deep-equal` with `react-fast-compare`

khmm12:
`react-fast-compare` is more lightweight and faster alternative.

### Demo app bundle size comparison

| Before    | After     | Diff     |
|-----------|-----------|----------|
| 288.67 KB | 255.96 KB | -32.71 KB |

### Performance comparison

[Benchmark](https://github.com/FormidableLabs/react-fast-compare/blob/master/benchmark/index.js) with included `deep-equal` (strict-mode) suite.

```
react-fast-compare x 196,844 ops/sec ±0.41% (92 runs sampled)
fast-deep-equal x 192,322 ops/sec ±0.23% (92 runs sampled)
lodash.isEqual x 34,136 ops/sec ±0.37% (94 runs sampled)
nano-equal x 156,040 ops/sec ±0.56% (90 runs sampled)
shallow-equal-fuzzy x 121,470 ops/sec ±1.42% (93 runs sampled)
deep-equal (strict) x 145 ops/sec ±1.77% (57 runs sampled)
```

196,844 ops/sec vs 145 ops/sec

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,8 +66,8 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.9.2",
-    "deep-equal": "^2.0.1",
     "prop-types": "^15.6.1",
+    "react-fast-compare": "^3.0.1",
     "warning": "^4.0.2"
   },
   "devDependencies": {
--- a/src/Popper.js
+++ b/src/Popper.js
@@ -1,5 +1,5 @@
 // @flow
-import deepEqual from 'deep-equal';
+import deepEqual from 'react-fast-compare';
 import * as React from 'react';
 import { createPopper } from '@popperjs/core';
 import {
@@ -206,7 +206,7 @@ export class InnerPopper extends React.Component<PopperProps, PopperState> {
     if (
       this.props.placement !== prevProps.placement ||
       this.props.strategy !== prevProps.strategy ||
-      !deepEqual(this.props.modifiers, prevProps.modifiers, { strict: true })
+      !deepEqual(this.props.modifiers, prevProps.modifiers)
     ) {
       // develop only check that modifiers isn't being updated needlessly
       if (process.env.NODE_ENV === 'development') {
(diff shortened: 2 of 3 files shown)


khmm12:
Just my opinion and nothing more. 

Personally I would:

1. Replace deep equal comparison with shallow.
2. Add to README a section describing circular updates issue and how to avoid.

by next reasons:

1. Shallow equal comparison for modifiers covers most of use cases.
2. Deep equal comparison solves circular updates issue until modifiers doesn't contain inline functions. That means deep equal didn't solve the issue, but "hidden" for only one case.

```jsx
const [, forceUpdate] = useReducer(v => v + 1, 0)

<Popper
   modifiers={[
     { name: 'boom', enabled: true, phase: 'write', fn: () => forceUpdate() } 
  ]}
>
/* code */
</Popper>
```

khmm12:
Ooops, I see react-popper is almost rewritten with hooks API and without `deep-equal` usage. So if the PR is not more actual, please close.

FezVrasta:
Thanks for the PR.

As you noticed, I removed the deep-equal dependency from the new version, but the circular update issue could happen with my pending PR. I think I'd rather defer the responsibility to the consumers, but I need to write some good docs to help consumers understand the issue.

FezVrasta:
I played a bit more with the new `usePopper` Hook, and I realized it's basically impossible to get it right at first, because as soon you pass an options object, it will start looping.

We should ask the consumers to always use `useMemo` to memoize the options, but it would be a terrible ergonomic... So I ended up adding react-fast-compare to avoid this issue.

khmm12:
@FezVrasta It's ok (or expected) due to reconciler implementation, because hook arguments are not external dependencies unlike of component props produce causes child component update without affect to external (parent). 

Taking into account that `react-fast-compare` is fast enough to not add extra noticeable overhead, it wouldn't be problem.

> We should ask the consumers to always use useMemo to memoize the options, but it would be a terrible ergonomic... 

IMHO options memoization should be required only if it's under developer responsibility like in circular update issue due to modifiers that produce component update. In other words implementation shouldn't affect to API.  So please don't change your mind, this is right compromise 🙏.

PR #352, opened 2020-04-14
fix: TypeScript types broken with latest popperjs-core

bengry:
Fixes https://github.com/popperjs/react-popper/issues/351

### DIFF
--- a/src/Popper.js
+++ b/src/Popper.js
@@ -5,6 +5,7 @@ import {
   type Placement,
   type PositioningStrategy,
   type VirtualElement,
+  type StrictModifiers,
   type Modifier,
 } from '@popperjs/core/lib';
 import { ManagerReferenceNodeContext } from './Manager';
@@ -13,7 +14,7 @@ import { unwrapArray, setRef } from './utils';
 import { usePopper } from './usePopper';
 
 type ReferenceElement = ?(VirtualElement | HTMLElement);
-type Modifiers = Array<$Shape<Modifier<any>>>;
+type Modifiers = Array<StrictModifiers | $Shape<Modifier<string, {}>>>;
 
 export type PopperArrowProps = {|
   ref: Ref,
--- a/typings/react-popper.d.ts
+++ b/typings/react-popper.d.ts
@@ -68,7 +68,7 @@ export class Popper<Modifiers> extends React.Component<
 > {}
 
 export function usePopper(
-  referenceElement?: Element | null,
+  referenceElement?: Element | PopperJS.VirtualElement | null,
   popperElement?: HTMLElement | null,
   options?: Partial<PopperJS.Options> & {
     createPopper?: typeof PopperJS.createPopper;
--- a/typings/react-popper.d.ts
+++ b/typings/react-popper.d.ts
@@ -67,11 +67,12 @@ export class Popper<Modifiers> extends React.Component<
   {}
 > {}
 
-export function usePopper(
+export function usePopper<Modifiers>(
   referenceElement?: Element | PopperJS.VirtualElement | null,
   popperElement?: HTMLElement | null,
-  options?: Partial<PopperJS.Options> & {
+  options?: Omit<Partial<PopperJS.Options>, 'modifiers'> & {
     createPopper?: typeof PopperJS.createPopper;
+    modifiers?: ReadonlyArray<Modifier<Modifiers>>;
   }
 ): {
   styles: { [key: string]: React.CSSProperties };
--- a/typings/react-popper.d.ts
+++ b/typings/react-popper.d.ts
@@ -1,5 +1,8 @@
-import * as React from 'react';
 import * as PopperJS from '@popperjs/core';
+import * as React from 'react';
+
+// Utility type
+type UnionWhere<U, M> = U extends M ? U : never;
 
 interface ManagerProps {
   children: React.ReactNode;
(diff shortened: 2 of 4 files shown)


src/Popper.js
```diff
@@ -13,7 +14,7 @@ import { unwrapArray, setRef } from './utils';
 import { usePopper } from './usePopper';
 
 type ReferenceElement = ?(VirtualElement | HTMLElement);
-type Modifiers = Array<$Shape<Modifier<any>>>;
+type Modifiers = Array<$Shape<StrictModifiers | Modifier<string, {}>>>;
```
bengry, line 17: I'm not very familiar with Flow's types, so the type here may be incorrect.
What I was going for is the equivalent of this in TypeScript:
```ts
type Modifiers = Array<StrictModifiers | Partial<Modifier<string, object>>>;
```
In simple terms - an array of items, where each item is either `StrictModifiers` (which is a union of types, so one of them), or `Partial<Modifier<string, object>>` - which is an object with the same shape as `Modifier`, in which all properties are optional - the generics passed to it are `string` for `Name` and any object (so `{ foo: 'bar' }` would work, but `true` wouldn't) for `Options` (note that `object` is not the same as the `Object` type).

atomiks: Do you know why autocomplete works but not typechecking? When I hover the tooltip it shows the types, but doesn't actually error for wrong types.

autocomplete

no error

atomiks: Okay, it seems like adding some dynamic behavior inside the modifiers array stops TypeScript from working 😅

bengry: I'm guessing this is a typescript file? Not sure from the screenshot.
Originally my changes weren't able to make the type-checking work properly, since a Modifer can be a Strict one (i.e. `name` is one of `	"offset" | "applyStyles" | "arrow" | "hide" | "computeStyles" | "eventListeners" | "flip" | "preventOverflow" | "popperOffsets"` or an external one (`name` is any string other than the aforementioned ones).

I've just pushed a change that should make the distinction at compile-time, and allow both IDE autocompletion as well as compile-time type checking for the different modifiers - so this wouldn't compile:
```ts
{
  name: 'flip',
  options: {
	altBoundary: 1  // Type 'number' is not assignable to type 'boolean | undefined'.
  }
}
```

and neither would this:
```ts
{
  name: 'offset',
  options: {
	unknownOption: true
  }
}
```

This would though:
```ts
{
  name: 'my-modifier',
  options: {
    someOption: true,
    anotherOption: 'test'
  }
}
```

FezVrasta:
Thanks for the PR! I have not a lot of experience with TS so I'll wait for @atomiks's go ahead before I proceed to merge it.

typings/react-popper.d.ts
```diff
@@ -37,22 +40,37 @@ export interface PopperChildrenProps {
...
+
+type Modifier<Name> = Name extends StrictModifierNames
+  ? StrictModifier<Name>
+  : Partial<PopperJS.Modifier<Name, object>>;
+
+export interface PopperProps<Modifiers> {
   children: (props: PopperChildrenProps) => React.ReactNode;
   innerRef?: React.Ref<any>;
-  modifiers?: Array<Partial<PopperJS.Modifier<any>>>;
+  modifiers?: ReadonlyArray<Modifier<Modifiers>>;
```
atomiks, line 57:
```suggestion
  modifiers?: ReadonlyArray<Modifiers>;
```

atomiks:
Works well with the change. I'd like to have this as a default in core too (typed built-in modifiers, untyped custom) but I don't know how the Flow conversion would handle this.

typings/react-popper.d.ts
```diff
@@ -37,22 +40,37 @@ export interface PopperChildrenProps {
...
 
-export function usePopper(
-  referenceElement?: Element | null,
+export function usePopper<Modifiers>(
+  referenceElement?: Element | PopperJS.VirtualElement | null,
   popperElement?: HTMLElement | null,
-  options?: Partial<PopperJS.Options> & {
+  options?: Omit<Partial<PopperJS.Options>, 'modifiers'> & {
     createPopper?: typeof PopperJS.createPopper;
+    modifiers?: ReadonlyArray<Modifier<Modifiers>>;
```
atomiks, line 73:
```suggestion
    modifiers?: ReadonlyArray<Modifiers>;
```

FezVrasta:
> Works well with the change. I'd like to have this as a default in core too (typed built-in modifiers, untyped custom) but I don't know how the Flow conversion would handle this.

I'm preparing a PR to update the Flow types of react-popper, and it seems to work just like TS.

atomiks:
My suggestions were because I wasn't inlining the modifiers array lol 🤦  — your change is fine.

How would I type this with the modifiers array assigned to a variable instead of inlined?

atomiks:
Is it possible to type custom modifiers with this? I'd say it's an important feature

typings/tests/main-test.tsx
```diff
@@ -60,7 +60,12 @@ const HookTest = () => {
     referenceElement,
     popperElement,
     {
-      modifiers: [{ name: 'arrow', options: { element: arrowElement } }],
+      modifiers: [
+        {
+          name: 'arrow',
+          options: { element: arrowElement as HTMLElement | undefined },
```
FezVrasta, line 66: I just released `@popperjs/core@2.3.3` to allow the arrow element to be `null`, you should be able to get rid of this `as` thingy

bengry:
> Is it possible to type custom modifiers with this? I'd say it's an important feature

Sure you can:
```ts
const modifiers: (StrictModifier | Modifier<'my-modifier'>)[] = [
  {
    name: 'offset',
    options: {},
  },
  {
    name: 'my-modifier',
    options: {
      someProperty: 'foo',
    },
  },
];
```

Unfortunately you can't omit the `my-modifier` in line 1 here (and use `string` instead of example) since then TS will collapse the two types to 		a union of all modifiers, and you lose type-safety for strict modifiers - so this will compile, even though it shouldn't (when used inside of an array typed as `(StrictModifier | Modifier<string>)[]`):
```ts
{
  name: 'offset'
  options: {
    adaptive: true,
  },
}
```

I usually prefer inline types for this reason, TypeScript is much more powerful when you let it infer types than when you explicitly tell the compiler what you want.

Maybe there's a workaround here that I'm missing (apart from creating two arrays - `const strictModifiers: StrictModifier[] = [ ... ]` and `const otherModifiers: Modifier<string>[] = [ ... ]` and then merging them later: `[...strictModifiers, ...otherModifiers]`).

atomiks:
I mean type custom modifers' options like the strict ones (inlined or not)

bengry:
> I mean type custom modifiers' options like the strict ones (inlined or not)

You can type them like that, they should be able to pass just fine to the `modifiers` prop/option. Are you having an issue with such a case?

atomiks:
Seems good now ✅

bengry:
@atomiks great, thanks! :)
@FezVrasta can we get this merged and released soon-ish please?

l0gicgate:
When will this be merged?

kybradeck:
Hi! Thanks for fixing and merging this, was just about to report the issue. Do you know when this fix will be released on NPM?

PR #357, opened 2020-04-21
fix: tweak hooks usage in usePopper

nstepien:


src/usePopper.js
```diff
@@ -41,7 +41,7 @@ export const usePopper = (
   const [state, setState] = React.useState<State>({
     styles: {
       popper: {
-        position: optionsWithDefaults.strategy || 'absolute',
```
nstepien, line 44: `optionsWithDefaults` already defaults `strategy` to `'absolute'`, so we don't need to do another check for a falsy value here.

src/usePopper.js
```diff
@@ -68,14 +68,14 @@ export const usePopper = (
       },
       requires: ['computeStyles'],
     }),
-    [setState]
```
nstepien, line 71: While `setState` does change on each render, it actually doesn't invalidate the cache! As it's bound to the same context, and React is smart enough somehow.
You can test this with the following:
```js
useMemo(() => {
  console.log('changed?');
}, [setState]);
```

src/usePopper.js
```diff
@@ -68,14 +68,14 @@ export const usePopper = (
...
+    []
   );
 
   const popperOptions = React.useMemo(() => {
     const newOptions = {
       onFirstUpdate: optionsWithDefaults.onFirstUpdate,
-      placement: optionsWithDefaults.placement || 'bottom',
-      strategy: optionsWithDefaults.strategy || 'absolute',
```
nstepien, line 78: Same as above.

src/usePopper.js
```diff
@@ -98,35 +98,32 @@ export const usePopper = (
   ]);
 
   const popperInstanceRef = React.useRef();
-  const createPopper = React.useMemo(
```
nstepien, line 101: This was only used in one `useIsomorphicLayoutEffect`, so I figured we can just inline it.

src/usePopper.js
```diff
@@ -98,35 +98,32 @@ export const usePopper = (
...
-        referenceElement,
-        popperElement,
-        popperOptions
-      );
-
-      popperInstanceRef.current = popperInstance;
+    if (popperInstanceRef.current) {
+      popperInstanceRef.current.setOptions(popperOptions);
+    }
+  }, [popperOptions]);
```
nstepien, line 106: I've moved this `useIsomorphicLayoutEffect` before the other one.
I assume we only want to call `setOptions(...)` when the options have changed. Calling it immediately after creating the popper instance would be wasteful, right?

src/usePopper.js
```diff
@@ -98,35 +98,32 @@ export const usePopper = (
...
+
+  useIsomorphicLayoutEffect(() => {
+    if (referenceElement == null || popperElement == null) {
+      return;
```
nstepien, line 110: We can return early here, that way we don't schedule a cleanup function that is useless when we don't create a popper instance.

src/usePopper.js
```diff
@@ -98,35 +98,32 @@ export const usePopper = (
...
+    );
+
+    popperInstanceRef.current = popperInstance;
+
     return () => {
-      popperInstance != null && popperInstance.destroy();
+      popperInstance.destroy();
       popperInstanceRef.current = null;
     };
-    // eslint-disable-next-line react-hooks/exhaustive-deps
```
nstepien, line 122: `react-hooks/exhaustive-deps` unfortunately doesn't/cannot handle checking the dependency arrays for "custom" hooks.
Do we really have to avoid using `useLayoutEffect` in SSR env?

FezVrasta: Without, SSR customers would get a warning that would be very difficult to get rid of, so I don't think there are better alternatives.

FezVrasta:
Thanks for the PR! It looks great.

nstepien:
:shipit:

PR #367, opened 2020-06-04
Don't return a value in useEffect callback

gnapse:
Not sure how no one has had this issue yet, but it made my code fail all over the place when upgrading from v1 to v2. The console was filled with these:

> Warning: An effect function must not return anything besides a function, which is used for clean-up. You returned null. If your effect does not require clean up, return undefined (or nothing).

I tracked it down to this effect using an implicit-return arrow function for a callback.


FezVrasta:
Thanks!

PR #374, opened 2020-07-06
fix: Returns correct state when referenceElement becomes null

lmk123:
When I change `referenceElement` or `popperElement` to `null`, `usePopper` returns old state to me.

## Steps To Reproduce

1. open the reproduce link: https://codepen.io/lmk123/pen/zYrRzMM
2. click the 'Set referenceElement to null' button

Expected behavior: `state.placement` becomes `null`
Current behavior: `state.placement` is still `bottom`

### DIFF
--- a/src/usePopper.js
+++ b/src/usePopper.js
@@ -107,6 +107,10 @@ export const usePopper = (
 
   useIsomorphicLayoutEffect(() => {
     if (referenceElement == null || popperElement == null) {
+      if (popperInstanceRef.current) {
+        popperInstanceRef.current.destroy();
+        popperInstanceRef.current = null;
+      }
       return;
     }
 


lmk123:
This is because the previous effect is cleaned up **_after_** the re-render::
https://overreacted.io/a-complete-guide-to-useeffect/#so-what-about-cleanup
> Say props is {id: 10} on the first render, and {id: 20} on the second render.
> - React renders UI for {id: 20}.
> - The browser paints. We see the UI for {id: 20} on the screen.
> - **_React cleans up the effect for {id: 10}._**
> - React runs the effect for {id: 20}.

When `referenceElement` becomes `null`, we need destroy the previous popper instance immediately.

PR #376, opened 2020-07-15
Update peerDependencies in package.json

MichaelDeBoey:
https://twitter.com/SemanticRelease/status/1249795728253353986

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,8 +58,8 @@
     "popover"
   ],
   "peerDependencies": {
-    "@popperjs/core": "^2.0.0",
-    "react": "^16.8.0"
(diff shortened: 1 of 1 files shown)


FezVrasta:
https://github.com/jaydenseric/apollo-upload-client/pull/201#issuecomment-657916102

MichaelDeBoey:
@FezVrasta I think it's the right approach, as do a lot of other packages by now.

FezVrasta:
I think the way it is now it's okay. I want to have control over these matters when a new major of a dependency is published. Thanks for the feedback though

PR #386, opened 2020-09-16
added typing for RefHandler

o-faro:
fixes issues 373

### DIFF
--- a/typings/react-popper.d.ts
+++ b/typings/react-popper.d.ts
@@ -9,6 +9,8 @@ interface ManagerProps {
 }
 export class Manager extends React.Component<ManagerProps, {}> {}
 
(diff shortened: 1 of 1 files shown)


typings/react-popper.d.ts
```diff
@@ -9,6 +9,8 @@ interface ManagerProps {
 }
 export class Manager extends React.Component<ManagerProps, {}> {}
 
+type RefHandler = (ref: HTMLElement | null) => void;
```
FezVrasta, line 12: Don't you need to export it?

o-faro: yes, I do.

FezVrasta:
I need somebody with TS experience to review this, I don't know if this change is enough to fix the issue or not.

PR #388, opened 2020-10-06
Update react-popper.d.ts

fleck:
The current type doesn't show that the popper object may be undefined during the first render.

```ts
const { styles, attributes } = usePopper(...

console.log(attributes.popper);
```

when rendering the object is first undefined then during a second render it's populated:

image

This type update would reflect that fact in TS

image


fleck:
That was fast 🏎. Thanks for maintaining this library, just started using it this morning and it's great!

Issue #393, opened 2020-11-09
Arrow not positioned correctly until a scroll happens


skimi:
### Reproduction demo

`%` is the arrow in this example. It is at the left of the element at initialisation when it should be centered.
Only when a scroll starts does it fix the issue.

https://codesandbox.io/s/react-popper-v2x-issue-template-forked-2fj5t?file=/src/index.js

Codesandebox sometimes does something with its layout that triggers the repositioning of the arrow so hit the refresh of the browser view or go to it directly https://2fj5t.csb.app/ 

![react-popper-no](https://user-images.githubusercontent.com/10855180/98572538-0d738980-22b6-11eb-96ed-1e8e98b1c1fe.gif)

### Steps to reproduce the problem

1. Create a react-popper top or bottom with arrow
2. Arrow not centered

### What is the expected behavior?

Arrow should be centered immediately

### Any other comments?

If I bypass react-popper and use popperjs directly inside a useLayoutEffect it works as expected

### Packages versions

- Popper.js: 2.5.4
- react-popper: 2.2.4

skimi:
The issue is with the custom update modifier:
```
  updateStateModifier,
  { name: 'applyStyles', enabled: false },
```

`updateStateModifier` is not receiving the correct initial arrow styling so it's possible the issue is in poperjs/core.
The old components Manager, Target, etc... does not create a custom updateStateModifier so it does not have the issue the hook has

Kinbaum:
@skimi @FezVrasta I still see this as an issue after updating to `2.5.5`.
The initialization of the arrow is now near the target element, but it's slightly to the right of the element. Also, where the content of the popper spawns isn't working.

In the below examples, I have my placement set to `bottom-end` and my arrow element has a padding of `12`. I have a modifier in place to set the offset to `[ 12, 12 ]`

Here is the initial arrow position:

<img width="211" alt="Screen Shot 2021-03-25 at 2 57 49 PM" src="https://user-images.githubusercontent.com/6590356/112528400-7ac21b80-8d7a-11eb-8e2e-ea59e6eadb94.png">

And here it is if scroll the page (the desired position):

<img width="185" alt="Screen Shot 2021-03-25 at 2 58 45 PM" src="https://user-images.githubusercontent.com/6590356/112528509-9b8a7100-8d7a-11eb-8b01-892fc815510d.png">

drewlustro:
How did you fix this @skimi ?

PR #394, opened 2020-11-10
fix: arrow not positioned properly at initialisation #393

skimi:
See #393 for description

I'll see about adding a test as soon as I can. Until then anyone having this issue can just add the "position: absolute" to the arrow via their own styling

### DIFF
--- a/src/usePopper.js
+++ b/src/usePopper.js
@@ -45,6 +45,9 @@ export const usePopper = (
         left: '0',
         top: '0',
       },
+      arrow: {
+        position: 'absolute', 
+      },
     },
     attributes: {},
   });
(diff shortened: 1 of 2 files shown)


FezVrasta:
Thanks! I'll wait for the test and then I'll merge it.

skimi:
@FezVrasta I've added a very basic test

I would have liked to actually check that the arrow is centered at initialization by checking the transform value returned by popperjs but did not manage it.
All positioning within jest are at 0 since there is no actual dom to do anything different. I tried mocking all the usual suspect like `getBoundingClientRect`, `offsetWidth` and `offsetHeight` but it wasn't enough 😞

Anyhow since it was too complex and knowing the real tool for this would be Cypress I kept it simple

FezVrasta:
Thanks!

Issue #405, opened 2021-02-23
Update dependencies in v1 to support react v17


h3rmanj:
Installing `react-popper@1` in a project containing `react@17`, with `npm@7` results in the following error:

```
> npm i react-popper@1
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! Found: react@17.0.1
npm ERR! node_modules/react
npm ERR!   peer react@"17.0.1" from react-dom@17.0.1
npm ERR!   node_modules/react-dom
npm ERR!     react-dom@"^17.0.1" from the root project
npm ERR!   react@"^17.0.1" from the root project
npm ERR!   1 more (react-popper)
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer react@"^0.14.0 || ^15.0.0 || ^16.0.0" from create-react-context@0.3.0
npm ERR! node_modules/react-popper/node_modules/create-react-context
npm ERR!   create-react-context@"^0.3.0" from react-popper@1.3.9
npm ERR!   node_modules/react-popper
npm ERR!     react-popper@"1" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
```

While there are active pull requests to create-react-context to support v17, it seems to be abandoned. A fork has been made to support it, and uploaded to npm as `@hypnosphi/create-react-context`. https://github.com/jamiebuilds/create-react-context/pull/32#issuecomment-758158196

I hope react-popper can use this package instead. Will add a merge request.

h3rmanj:
Keeping this one open until it's released 🙂

FezVrasta:
1.3.10 includes this fix :-)

h3rmanj:
Oh sorry, I was too quick. Thanks! 😄

PR #406, opened 2021-02-23
use @hypnosphi/create-react-context to support react 17 in v1

h3rmanj:
Addresses #405 

In short: `create-react-context` doesn't support react v17, and since it looks to be abandoned, use `@hypnosphi/create-react-context` instead.


FezVrasta:
Thanks!

FezVrasta:
1.3.10 is live with this fix

PR #425, opened 2021-06-29
Only set Reference's innerRef to null when unmounting

shepmaster:
Without this, the reference is set to null for _every render_, making
it mostly unusable.


FezVrasta:
Thanks!

Agontuk:
@FezVrasta can you please publish a new version with this change ? Faced an issue regarding this, then realized this fix was not yet published.

PR #433, opened 2021-12-28
Export usePopper options type

valtism:
Exporting the options type is very useful as it is currently impossible to get it using `Parameters<typeof usePopper>[2]`  as extending an interface in Typescript will throw the error `An interface can only extend an identifier/qualified-name with optional type arguments` because of the `<TModifier extends Partial` inside `createPopper`'s type.


henrijeret:
Please merge this :)

